PR #3, opened 2017-12-13
Managed service issue history

Dorthu:
This is migrating https://github.com/linode/manager/pull/2697


src/data/endpoints/managed.yaml
```diff
@@ -81,3 +80,201 @@ endpoints:
...
+          address:
+            description: The URL to request. If service_type is TCP, must include port number. For example, tcp://monitor.example.org:23
+            type: String
+            value: https://monitor.example.org
+          timeout:
+            description: Response must be received before this timeout or an alert will be triggered.
+            value: 30
+            limit: 1-255
+          body:
+            description: Response must contain this string or an alert will be triggered.
```
SzySteve: body is optional according to the blueprint

src/data/endpoints/managed.yaml
```diff
@@ -81,3 +80,201 @@ endpoints:
...
+            type: String
+            value: Success!
+            limit: 0-100 chracters
+          consultation_group:
```
SzySteve: optional

src/data/endpoints/managed.yaml
```diff
@@ -81,3 +80,201 @@ endpoints:
...
+            description: If we need help along the way, we'll contact someone from this group.
+            value: Support
+            limit: 0-50 characters
+          region:
+            description: Required if address is a private IP, otherwise not used.
```
SzySteve: technically optional

src/data/endpoints/managed.yaml
```diff
@@ -81,3 +80,201 @@ endpoints:
...
+            type: String
+            value: us-east-1a
+          notes:
+            description: We'll refer to these instructions when responding to a service-related issue.
```
SzySteve: optional

SzySteve: wait do we even mark these as optional in the endpoints yaml?

PR #6, opened 2017-12-14
Document rendered zone file

Dorthu:


src/data/objects/domain.yaml
```diff
@@ -107,6 +107,25 @@ schema:
     filterable: true
     subtype: domain_type
     description: Controls the Domain type.
+  zonefile:
+    type: Object
+    description: >
+      Information about the rendered zone file.  This shows what is live right
+      now, and recent changes may take a minute to be reflected in this value.
```
yoshizzle: Is it an actual minute? I thought it could take up to fifteen minutes?

If it's more than an actual minute, we should specify with "may take up to ..." kind of message.

src/data/objects/domain.yaml
```diff
@@ -107,6 +107,25 @@ schema:
...
+    rendered:
+      type: String
+      description: >
+        The rendered zone file as it is right now.  This value is refreshed
+        periodically, and may take a minute to reflect changes.
```
yoshizzle: same

PR #8, opened 2017-12-15
ref: cleanup managed credential docs

SzySteve:
* remove username and last_decrypted from the response
* move /:id/password to /:id/update

screen shot 2017-12-15 at 11 37 41 am
screen shot 2017-12-15 at 11 38 03 am

### DIFF
--- a/src/data/endpoints/managed.yaml
+++ b/src/data/endpoints/managed.yaml
@@ -252,12 +252,12 @@ endpoints:
     type: resource
     authenticated: true
     description: >
-      Change the username and password for a credential.
+      Change the username and/or password for a credential.
     methods:
       POST:
         response: ManagedCredential
         description: >
-          Changes the username and password for a credential.
+          Changes the username and/or password for a credential.
         params:
           username:
             optional: True
--- a/src/data/endpoints/managed.yaml
+++ b/src/data/endpoints/managed.yaml
@@ -247,22 +247,33 @@ endpoints:
                     "username": "myusername2",
                 }' \
                 https://$api_root/$version/managed/credentials/$credential_id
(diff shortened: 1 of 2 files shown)


src/data/endpoints/managed.yaml
```diff
@@ -247,22 +247,33 @@ endpoints:
...
                 }' \
                 https://$api_root/$version/managed/credentials/$credential_id
-  /managed/credentials/$id/password:
+  /managed/credentials/$id/update:
     group: Credentials
     type: resource
     authenticated: true
     description: >
-      Change the password for a credential.
+      Change the username and password for a credential.
```
Dorthu: you can do one or both, right?  Should this read "username and/or password"?

src/data/endpoints/managed.yaml
```diff
@@ -247,22 +247,33 @@ endpoints:
...
     methods:
       POST:
         response: ManagedCredential
         description: >
-          Changes the password for a credential.
+          Changes the username and password for a credential.
```
Dorthu: same here, should this be "and/or" ?

src/data/endpoints/managed.yaml
```diff
@@ -247,22 +247,33 @@ endpoints:
...
+        params:
+          username:
+            optional: True
+            description: New username to set for this credential.
+            value: myusername
+            type: String
+          password:
+            description: New password to set for this credential.
+            value: mypassword
+            type: String
```
Dorthu: password is required but username isn't?  That's ...weird 😕

yoshizzle: Weird, but correct.

PR #15, opened 2017-12-19
add managed issues to docs

SzySteve:
Add managed issues to docs


src/data/endpoints/managed.yaml
```diff
@@ -89,21 +89,34 @@ endpoints:
...
+  /managed/issues:
+    group: Issues
+    authenticated: true
+    description: >
+      A list of the issues created from monitoring your Services.
+    methods:
+      GET:
+        response: ManagedIssue
+        description: >
+          Returns your managed service issue history list.
```
Dorthu: I would remove the word "history" from here - this is a collection of your managed issues

src/data/objects/managedissue.yaml
```diff
@@ -0,0 +1,53 @@
...
+description: >
+  Managed Issue objects describe support ticket issues that have occurred while subscribed to your Managed service.
+schema:
+  created:
+    type: Datetime
+    value: 2017-10-31T11:12:21
+    description: Time at which the issue was reported.
+  entity:
+    description: >
+      Displays information related to the type of issue being monitored by your Service.
```
Dorthu: description should not be inside of entity here

src/data/objects/managedissue.yaml
```diff
@@ -0,0 +1,53 @@
...
+      type: String
+      value: /support/tickets/1234567
+      description: >
+        The endpoint URI related to the Support Ticket that was created.
+  services:
+    type: Array
+    description: >
+      List of impacted service IDs.
+    seeAlso: ["/endpoints/managed/services"]
+    value: [111, 123]
```
Dorthu: missing "id" attribute (outside of entity)

PR #16, opened 2017-12-19
add credentials as a param for managed services

SzySteve:


src/data/endpoints/managed.yaml
```diff
@@ -158,6 +158,15 @@ endpoints:
...
             limit: 0-1000 characters
             optional: true
+          credentials:
+            type: Array
+            subtype: Integer
+            value: [12345, 23456, 34567]
+            editable: true
+            optional: true
+            description: >
+              An array of credentials to associate with this service.
```
Dorthu: An array of credential IDs to associate?

src/data/objects/managedservice.yaml
```diff
@@ -69,16 +69,14 @@ schema:
...
-  #  description: >
-  #    An array of credentials associated with this service.
+  credentials:
+    type: Array
+    subtype: Integer
+    value: [12345, 23456, 34567]
+    editable: true
+    optional: true
+    description: >
+      An array of credentials associated with this service.
```
Dorthu: credential IDs?

PR #17, opened 2017-12-19
merge conflict got through into develop

SzySteve:
hopefully this just puts us in a sane state again. This is fine in master.


Dorthu:
sorry, this was probably my bad

PR #18, opened 2017-12-19
Adds object and endpoints for GET and PUT /managed/linode_settings

yoshizzle:


src/data/objects/managedlinodesetting.yaml
```diff
@@ -0,0 +1,51 @@
...
+      editable: true
+      value: root
+      description: >
+        The SSH username used to log into your Linode.
+    ip:
+      type: String
+      editable: true
+      value: 204.23.61.122
+      description: >
+        One of "any", IPv4, or IPv6 address to use to SSH into the Linode.
```
SzySteve: This phrasing feels weird. Maybe
`Either "any" or a valid IP address. Both IPv4 and IPv6 are supported`
or something like that. Lets ask the team

yoshizzle: They can't be any old valid IP address. They have to be associated with the Linode specifically.

SzySteve: Lets call that out too then

PR #21, opened 2017-12-20
Docs/feedback cleanup

SzySteve:
Addressing user feedback around docs inaccuracies

### DIFF
--- a/src/data/objects/image.yaml
+++ b/src/data/objects/image.yaml
@@ -39,7 +39,6 @@ schema:
   is_public:
     type: Boolean
     value: false
-    filterable: true
     description: True if the image is public.
   size:
     type: Integer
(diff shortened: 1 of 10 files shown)


src/data/objects/image.yaml
```diff
@@ -39,7 +39,6 @@ schema:
   is_public:
     type: Boolean
     value: false
-    filterable: true
```
yoshizzle: is_public is filterable per https://bits.linode.com/LinodeAPI/apinext/blob/development/apinext/maps/image.py#L53

src/data/objects/profile.yaml
```diff
@@ -50,7 +50,7 @@ schema:
       value: 2
       description: Total number of referrals attributed to user that have not yet converted to full accounts.
     credit:
-      type: Integer
+      type: Float
       value: 160
```
yoshizzle: if we're doing 40.0 as the example for longviewsubscription.yaml, shouldn't we also add the ".0" at the end of all of these other US dollar float value examples as well?

src/data/objects/image.yaml
```diff
@@ -39,7 +39,7 @@ schema:
...
+    fitlerable: true
```
jamesottinger: ^ sp

PR #22, opened 2017-12-21
new: Added endpoint/obj for managed stat, modified prebuild

yoshizzle:
Prebuild was not able to properly recurse through objects if they didn't
contain sub-objects.


yoshizzle:
Would probably be a good idea to get @jamesottinger's buy-in on this before merging.

PR #23, opened 2017-12-21
new: Added PUT /managed/services/$id

yoshizzle:


src/data/endpoints/managed.yaml
```diff
@@ -211,6 +211,25 @@ endpoints:
...
+          curl: |
+            curl -H "Content-Type: application/json" \
+                -H "Authorization: Bearer $TOKEN" \
+                -X PUT -d '{
+                    "label": "NewerMonitor",
+                    "service_type": "url",
+                    "address": "http://monitor2.example.org",
+                    "timeout": 60,
+                    "consultation_group": "DevOps",
+                    "region": "us-east-1a",
```
SzySteve: can region be changed?

yoshizzle: yes, if it's a private IP for address

jamesottinger: region should be null in this example

yoshizzle:
added change @jamesottinger asked for regarding null region in the example

PR #31, opened 2018-01-03
Added more robust language to kvmify

yoshizzle:
More explicit language denoting that you can only kvmify if your Linode
is currently Xen.

### DIFF
--- a/src/data/endpoints/linodes.yaml
+++ b/src/data/endpoints/linodes.yaml
@@ -655,8 +655,11 @@ endpoints:
     type: Action
     authenticated: true
     description: >
-      Changes a Linode's hypervisor from Xen to KVM. This endpoint can only be
-      used if your Linode is currently Xen and you wish to move to KVM.
+      Changes a Linode's hypervisor from Xen (legacy) to KVM (modern). Doing this
+      restarts and migrates your Linode, and can take several minutes depending on
+      the size of the Linode. Upgrading to KVM has significant performance
+      improvements. This endpoint will only work for Linodes currently running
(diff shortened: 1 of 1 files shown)


src/data/endpoints/linodes.yaml
```diff
@@ -655,7 +655,8 @@ endpoints:
     type: Action
     authenticated: true
     description: >
-      Changes a Linode's hypervisor from Xen to KVM.
+      Changes a Linode's hypervisor from Xen to KVM. This endpoint can only be
+      used if your Linode is currently Xen and you wish to move to KVM.
```
Dorthu: Might be worth including more info on what KVM and Xen are, and why the upgrade is desirable?  Something like 

> Changes a Linode's hypervisor from Xen (legacy) to KVM (modern).  Doing this restarts and migrates the Linode, and can take several minutes depending on the size of the Linode.  Upgrading to KVM has significant performances improvements, including a 300% boost in Unixbench scores and faster boots and shutdowns.  This endpoint will only work for Linodes running on the Xen hypervisor.

src/data/endpoints/linodes.yaml
```diff
@@ -655,7 +655,11 @@ endpoints:
     type: Action
     authenticated: true
     description: >
-      Changes a Linode's hypervisor from Xen to KVM.
+      Changes a Linode's hypervisor from Xen (legacy) to KVM (modern). Doing this
+      restarts and migrates your Linode, and take several minutes depending on
```
jamesottinger: and can take

PR #34, opened 2018-01-04
Adds managed/services/$id/enable and disable endpoint documentation

yoshizzle:


src/data/endpoints/managed.yaml
```diff
@@ -238,6 +238,36 @@ endpoints:
...
+  /managed/services/$id/disable:
+    group: Services
+    type: Action
+    authenticated: true
+    description: >
+      Disables a Managed service your account may access.
+    methods:
+      POST:
+        description: >
+          Disables this Managed service.
```
Dorthu: maybe add some details about what disabling a service does?  I assume disabled services aren't monitored - you should say so

src/data/endpoints/managed.yaml
```diff
@@ -238,6 +238,36 @@ endpoints:
...
+  /managed/services/$id/enable:
+    group: Services
+    type: Action
+    authenticated: true
+    description: >
+      Enables a Managed service your account may access.
+    methods:
+      POST:
+        description: >
+          Enables this Managed service.
```
Dorthu: You should add what this actually means - enabled services are actively monitored, and maybe that this only really does anything if the service in question is presently disabled.

src/data/endpoints/managed.yaml
```diff
@@ -238,6 +238,39 @@ endpoints:
...
+    group: Services
+    type: Action
+    authenticated: true
+    description: >
+      Disables a Managed service your account may access.
+    methods:
+      POST:
+        description: >
+          Disables this Managed service. Will only disabled if it is currently
+          enabled. A disabled service will not be actively monitored.
```
Dorthu: Maybe link to [this guide](https://www.linode.com/docs/platform/linode-managed/) that has info on disabling a service/steal words from it.  Of note, you say `Will only disabled if` above.

PR #37, opened 2018-01-05
Added descriptive language to what Linode rebuild returns

yoshizzle:
It returns disks and configs JSON.


src/data/endpoints/linodes.yaml
```diff
@@ -1019,7 +1019,7 @@ endpoints:
     description: >
       Deletes all Disks and Configs on this Linode, then deploys a new
       Distribution or Image to this Linode with the given attributes. Returns
-      information about this Linode.
+      a JSON object representation of the Linode's disks and disk configs.
```
Dorthu: "Linode's disks and configs", no ?

yoshizzle: I mean, yeah?

Easy enough to change. Two secs...

PR #44, opened 2018-01-12
Rework the account/settings endpoint

Dorthu:
This is a proposal for how the new endpoints will look.  Please review
and comment here so I can implement.

### DIFF
--- a/src/data/endpoints/account.yaml
+++ b/src/data/endpoints/account.yaml
@@ -16,15 +16,15 @@ endpoints:
           Returns your account settings.
         examples:
           curl: |
-            curl -h "Authorization: Bearer $TOKEN" \
+            curl -H "Authorization: Bearer $TOKEN" \
                 https://$api_root/$version/account
       PUT:
         description: >
           Edits your account settings.
         examples:
           curl: |
-            curl -h "content-type: application/json" \
-                -h "authorization: bearer $token" \
+            curl -H "Content-Type: application/json" \
+                -H "Authorization: Bearer $TOKEN" \
                 -X PUT -d '{
                   "address_1": "123 Main St.",
                   "address_2": "Suite 101",
@@ -52,15 +52,15 @@ endpoints:
           Returns your current defaults.
         examples:
           curl: |
-            curl -h "Authorization: Bearer $TOKEN" \
+            curl -H "Authorization: Bearer $TOKEN" \
                 https://$api_root/$version/account/settings
       PUT:
         description: >
           Edits your account's defaults.
         examples:
           curl: |
-            curl -h "Content-Type: application/json" \
-                -h "Authorization: Bearer $TOKEn" \
+            curl -H "Content-Type: application/json" \
+                -H "Authorization: Bearer $TOKEN" \
                 -X PUT -d '{
                   "network_helper": true,
                   "longview_subscription": "longview-10"
--- a/src/data/endpoints/account.yaml
+++ b/src/data/endpoints/account.yaml
@@ -5,11 +5,10 @@ description: >
   Account endpoints provide a means of manipulating your account, including users,
   OAuth Clients and Tokens, and Events.
 endpoints:
-  /account/settings:
-    group: Settings
+  /account:
     type: resource
     description: >
-      Manage your account settings.
+        Managed account contact and billing information.
     methods:
       GET:
         response: account
@@ -17,16 +16,15 @@ endpoints:
           Returns your account settings.
         examples:
           curl: |
-            curl https://$api_root/$version/account/settings
-          python: |
-            my_settings = client.account.get_settings()
+            curl -h "Authorization: Bearer $TOKEN" \
+                https://$api_root/$version/account
       PUT:
         description: >
           Edits your account settings.
         examples:
           curl: |
-            curl -H "Content-Type: application/json" \
-                -H "Authorization: Bearer $TOKEN" \
+            curl -h "content-type: application/json" \
+                -h "authorization: bearer $token" \
                 -X PUT -d '{
                   "address_1": "123 Main St.",
                   "address_2": "Suite 101",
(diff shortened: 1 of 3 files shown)


src/data/endpoints/account.yaml
```diff
@@ -36,23 +34,39 @@ endpoints:
...
+          curl: |
+            curl -h "Authorization: Bearer $TOKEN" \
+                https://$api_root/$version/account/settings
+      PUT:
+        description: >
+          Edits your account's defaults.
+        examples:
+          curl: |
+            curl -h "Content-Type: application/json" \
+                -h "Authorization: Bearer $TOKEn" \
```
yoshizzle: capital N in TOKEN, please

src/data/objects/account.yaml
```diff
@@ -16,10 +27,14 @@ schema:
     type: String
     description: City portion of this account's billing address.
     value: Philadelphia
-  company:
+  state:
     type: String
-    description: This account's company name.
-    value: My Company LLC
+    description: State (or province) portion of this account's billing address.
```
yoshizzle: Might be helpful to explain in detail that if the customer is not in the US, please enter the province here. I know you have (or province) in parens, but I know from experience dealing with a lot of Asian folks - for example - that it's not entirely clear.

That one Thai customer account from two months ago comes to mind as well.

I guess I'm saying, let's add a little more detail in the description to be 100% perfectly clear.

src/data/objects/account.yaml
```diff
@@ -16,10 +27,14 @@ schema:
     type: String
     description: City portion of this account's billing address.
```
yoshizzle: do we need to say "portion"? Seems weird. Maybe:

The City for this account's billing address.

src/data/objects/account.yaml
```diff
@@ -16,10 +27,14 @@ schema:
...
-  company:
+  state:
     type: String
-    description: This account's company name.
-    value: My Company LLC
+    description: State (or province) portion of this account's billing address.
+    value: PA
+  zip:
+    type: String
+    description: Zip Code portion of this account's billing address.
```
yoshizzle: The Zip Code for this account's billing address.

src/data/endpoints/account.yaml
```diff
@@ -5,28 +5,26 @@ description: >
...
+                https://$api_root/$version/account
       PUT:
         description: >
           Edits your account settings.
         examples:
           curl: |
-            curl -H "Content-Type: application/json" \
-                -H "Authorization: Bearer $TOKEN" \
+            curl -h "content-type: application/json" \
+                -h "authorization: bearer $token" \
```
yoshizzle: Authorization
Bearer
$TOKEN

src/data/endpoints/account.yaml
```diff
@@ -36,23 +34,39 @@ endpoints:
...
+      PUT:
+        description: >
+          Edits your account's defaults.
+        examples:
+          curl: |
+            curl -h "Content-Type: application/json" \
+                -h "Authorization: Bearer $TOKEn" \
+                -X PUT -d '{
+                  "network_helper": true,
                   "longview_subscription": "longview-10"
```
yoshizzle: So there is nothing else to edit in account/settings now?

If there is, please add them. If not, then, uh... cool!

Dorthu: with this proposal, there is not

src/data/objects/account.yaml
```diff
@@ -14,12 +25,18 @@ schema:
...
-    description: City portion of this account's billing address.
+    description: The city for this account's billing address.
     value: Philadelphia
-  company:
+  state:
     type: String
-    description: This account's company name.
-    value: My Company LLC
+    description: >
+      State portion of this account's billing address, or providence if you
```
yoshizzle: province, not providence

src/data/endpoints/account.yaml
```diff
@@ -5,28 +5,26 @@ description: >
...
+            curl -h "Authorization: Bearer $TOKEN" \
+                https://$api_root/$version/account
       PUT:
         description: >
           Edits your account settings.
         examples:
           curl: |
-            curl -H "Content-Type: application/json" \
-                -H "Authorization: Bearer $TOKEN" \
+            curl -h "Content-Type: application/json" \
```
RobertDeRose: why did these get changed to `-h` from `-H` when the man for curl says `-H` is correct?

src/data/endpoints/account.yaml
```diff
@@ -36,23 +34,39 @@ endpoints:
...
+    description: >
+      Manage default settings for your account.
+    methods:
+      GET:
+        response: account_settings
+        description: >
+          Returns your current defaults.
+        examples:
+          curl: |
+            curl -h "Authorization: Bearer $TOKEN" \
```
RobertDeRose: same here

src/data/endpoints/account.yaml
```diff
@@ -36,23 +34,39 @@ endpoints:
...
+                https://$api_root/$version/account/settings
+      PUT:
+        description: >
+          Edits your account's defaults.
+        examples:
+          curl: |
+            curl -h "Content-Type: application/json" \
```
RobertDeRose: and here

src/data/endpoints/account.yaml
Dorthu: 😞  I think I hit `u` instead of `y` one time and didn't notice

src/data/objects/account.yaml
RobertDeRose: why you hating on Rhode Island?

src/data/endpoints/account.yaml
RobertDeRose: `-H` here too

src/data/endpoints/account.yaml
RobertDeRose: `-H` here as well

src/data/objects/account.yaml
yoshizzle: This is still not fixed

src/data/objects/account.yaml
```diff
@@ -1,9 +1,20 @@
-# vim: set et ts=2 sw=2 tw=80 :
 name: Account
-prefix: account/settings
+prefix: account
 description: >
-    A representation of account settings.
+    Your account contact and billing information.
 schema:
+  first_name:
```
RobertDeRose: Missing PUT output

src/data/endpoints/account.yaml
```diff
@@ -36,23 +34,39 @@ endpoints:
...
+                }' \
+                https://$api_root/$version/account
+  /account/settings:
+    group: Settings
+    type: resource
+    description: >
+      Manage default settings for your account.
+    methods:
+      GET:
+        response: account_settings
```
RobertDeRose: Should probably include something like `is_managed`

Dorthu: Done

PR #60, opened 2018-01-26
brk: Moved /linode/instances/$id/sharing to /networking/ip-sharing

yoshizzle:
* Endpoint is now /networking/ip-sharing
* linode_id is now a required field, in addition to current ips field


src/data/endpoints/networking.yaml
```diff
@@ -174,3 +174,32 @@ endpoints:
...
+      POST:
+        description: >
+            Sets IP Sharing for this Linode.
+        params:
+          linode_id:
+            type: Integer
+            description: >
+              The ID of the Linode to set up IP sharing for.
+          ips:
+            type: String
```
Dorthu: This is not a string.  This should be `type: Array` and `subtype: String` I think

PR #74, opened 2018-02-22
Added link to Linode CLI on docs main page

Dorthu:
I added the link in the dialog about the cURL guide, since it's another way to access the api.  If this placement doesn't make sense, please suggest an alternative.

Screenshot: 

screen shot 2018-02-22 at 1 19 59 pm


howdiz:
@Dorthu I suggest making the language more positive. 

Rather than it be about what the user does **not** want, we can just present an alternative option. 

I also suggest putting it on its own line.

Dorthu:
Reworded and moved to its own line:

screen shot 2018-03-02 at 7 29 55 am

PR #77, opened 2018-03-05
Fix get_regions() example

rmcintosh:
I also removed the last modified/byline subtitle

Closes https://github.com/linode/python-linode-api/issues/99


displague:
Printing the `id` would be more illustrative.

rmcintosh:
i'm not sure it matters

rmcintosh:
i should probably just copy the output

PR #103, opened 2019-08-23
[DO NOT MERGE] Example ssl cert and key + redactedInResponse and redactedString fields

nmelehan:
Added some example data for: ssl_cert, ssl_key, ssl_fingerprint,
and ssl_commonname

Introduced redactedInResponse and redactedString fields for the
ssl_cert and ssl_key properties. This is a proposed way of
communicating that a field should display <REDACTED> in the
response for a resource. This would require theme support.


nmelehan:
**DO NOT MERGE**

PR #106, opened 2019-08-26
[Update] DELETE /linode/instances/{linodeId}/disks/{diskId}

hzoppetti:
added a note that a disk can only be deleted if not attached to a linode with a currently running config profile

Documents:
-[ARB-1406] https://jira.linode.com/browse/ARB-1406

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4359,7 +4359,9 @@ paths:
       - Linode Instances
       summary: Delete Disk
       description: |
-        Deletes a Disk you have permission to `read_write`.
+        Deletes a Disk you have permission to `read_write`. The Disk cannot be
+        deleted if it is currently attached to a Linode with a currently-running
+        configuration profile.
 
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -4359,7 +4359,8 @@ paths:
       - Linode Instances
       summary: Delete Disk
       description: |
-        Deletes a Disk you have permission to `read_write`.
+        Deletes a Disk you have permission to `read_write`. The Disk cannot be
+        deleted if it is attached to a Linode's running configuration profile.
```
MichaelSaah: this might be clearer as
`The Disk cannot be deleted if it is attached to a running Linode. You can view a Linode's Disks by examining its configuration profile.`

MichaelSaah: On second thought, hold off on the above for a bit; clearing something up.

leslitagordita:
@MichaelSaah Okay, thanks for digging into this a bit more. Will close for now and standby for any upcoming changes.

PR #115, opened 2019-08-28
Update deploy/tagging script

nmelehan:
- Specify remote for push commands
- Link to GitHub new PR page

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -89,7 +89,7 @@ if git tag | grep -Fxq "${git_tag}"; then
 else
   echo "ERROR - ${git_tag} is not present in git tags"
   exit_status=1
-fi
+fis
 
 if [ $exit_status -ne 0 ]; then
   echo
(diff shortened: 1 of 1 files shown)


bin/deploy
```diff
@@ -100,6 +100,8 @@ else
   echo "Release process is complete."
   echo
   echo "Once confirmed changes are verified, you can:"
-  echo "  git push release-${version}"
-  echo "  git push ${git_tag}"
+  echo "  git push upstream release-${version}"
+  echo "  git push upstream ${git_tag}"
```
stvnjacobs: I would be against hard-coding the branch to push to. These are more of a suggestion, as not everyone has an `upstream` branch. Some keep `origin` as `origin` and others `upstream` or another convention.

stvnjacobs: Just documenting that you should be pushing to the Linode org repo should be enough, in my opinion.

PR #116, opened 2019-08-28
[Update] GET /account/events and GET /account/events/{eventId}

hzoppetti:
Added secondary_entity object to Event schema object which appears in both 
GET /account/events
and
GET /account/events/{eventId}

Additionally, while researching discovered that "filterable" should have been set on Event properties id and created - added that in second commit.

Documents:
- [ARB-1376] https://jira.linode.com/browse/ARB-1376
https://bits.linode.com/LinodeAPI/apinext/pull/1793
- [CT-753] https://jira.linode.com/browse/CT-753

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -12392,7 +12392,7 @@ components:
           format: date-time
           description: When this Event was created.
           example: '2018-01-01T00:01:01'
-          x-linode-cli-display: 5
+          x-linode-cli-display: 6
           x-linode-filterable: true
         entity:
           type: object
@@ -12455,7 +12455,7 @@ components:
               - volume
               readOnly: true
               description: >
-                The type of entity this Event is related to.
+                The type of entity that is being referenced by the Event.
               example: ticket
               x-linode-filterable: true
             url:
@@ -12470,7 +12470,8 @@ components:
           readOnly: true
           description: |
             Detailed information about the Event's secondary entity, which provides additional information
-            for Linode boot, create, and clone jobs
+            for `linode_boot`, `linode_create`, and `linode_clone` Event actions.
+          x-linode-cli-display: 5
           properties:
             id:
               type: integer
@@ -12479,9 +12480,6 @@ components:
 
 
                 Secondary Entities are assigned the ID of the Linode they correspond to.
-                When filtering by ID for these Events, use the corresponding Linode's ID.
-                These Events include:
-                  * `linode`
               example: 1234
             label:
               type: string
@@ -12491,9 +12489,13 @@ components:
               example: linode1234
             type:
               type: string
+              enum:
+              - backup
+              - image
+              - linode
               readOnly: true
               description: >
-                The type of entity this Event is related to.
+                The type of entity that is being referenced by the Event.
               example: linode
             url:
               type: string
@@ -12522,13 +12524,13 @@ components:
           readOnly: true
           description: If this Event has been read.
           example: true
-          x-linode-cli-display: 8
+          x-linode-cli-display: 9
         seen:
           type: boolean
           readOnly: true
           description: If this Event has been seen.
           example: true
-          x-linode-cli-display: 7
+          x-linode-cli-display: 8
         status:
           type: string
           readOnly: true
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -12463,6 +12465,45 @@ components:
                 a relative URL, it is relative to the domain you retrieved the
                 Event from.
               example: /v4/support/tickets/11111
+        secondary_entity:
+          type: object
+          readOnly: true
+          description: |
+            Detailed information about the Event's secondary entity, which provides additional information
+            for `linode_boot`, `linode_create`, and `linode_clone` Event actions.
+          x-linode-cli-display: 5
```
Dorthu: This is probably not going to display correctly; `entity` is listed as displaying in column 4, but in practice doesn't show up when you run `linode-cli events list`

openapi.yaml
```diff
@@ -12463,6 +12465,45 @@ components:
                 a relative URL, it is relative to the domain you retrieved the
                 Event from.
               example: /v4/support/tickets/11111
+        secondary_entity:
+          type: object
+          readOnly: true
+          description: |
+            Detailed information about the Event's secondary entity, which provides additional information
+            for `linode_boot`, `linode_create`, and `linode_clone` Event actions.
```
Dorthu: There will be others in the future.

openapi.yaml
```diff
@@ -12463,6 +12465,45 @@ components:
...
+          x-linode-cli-display: 5
+          properties:
+            id:
+              type: integer
+              description: |
+                The unique ID for an Event's secondary entity.
+
+
+                Secondary Entities are assigned the ID of the Linode they correspond to.
```
Dorthu: This isn't true; this is the ID of the object that is the secondary entity (i.e. if I deploy a Linode from the official Debain9 image, the secondary entity's ID will be `linode/debain9` because that's the ID of that image).  They may have nothing to do with a Linode.

openapi.yaml
```diff
@@ -12463,6 +12465,45 @@ components:
...
+        secondary_entity:
+          type: object
+          readOnly: true
+          description: |
+            Detailed information about the Event's secondary entity, which provides additional information
+            for `linode_boot`, `linode_create`, and `linode_clone` Event actions.
+          x-linode-cli-display: 5
+          properties:
+            id:
+              type: integer
```
Dorthu: Not always; images have slugs, for example

openapi.yaml
```diff
@@ -12463,6 +12465,45 @@ components:
...
+                The unique ID for an Event's secondary entity.
+
+
+                Secondary Entities are assigned the ID of the Linode they correspond to.
+              example: 1234
+            label:
+              type: string
+              description: >
+                The current label of this object. The label may reflect changes
+                that occur with this Event.
```
Dorthu: Actually no; secondary entity labels won't change.  Entity labels may stop being updated in the future, also.

openapi.yaml
```diff
@@ -12463,6 +12465,45 @@ components:
...
+              example: linode1234
+            type:
+              type: string
+              enum:
+              - backup
+              - image
+              - linode
```
Dorthu: This could be _any_ entity type.  Right now secondary entity specifically will only come back with "linode", "image", "linode_config", or "backup", but there may be more in the future.

openapi.yaml
hzoppetti: We noticed this as well and assumed this is because it's not displaying entity because it is an object? We assume this will be skipped as well but put it here in the case that should it in the future be displayed, that it will appear next to entity.

openapi.yaml
hzoppetti: Cool - I'll make this a regular string - Also - noticed that this was "backup" in this test but other places in the code this was was "backups"...

openapi.yaml
hzoppetti: Will update to be a string.

Dorthu: That's probably safer.  This could also come back as an int, but it doesn't look like openapi3 supports this.

PR #117, opened 2019-08-28
[Update] POST /linode/instances/{linodeId}/backups/{backupId}/restore

hzoppetti:
Updated POST /linode/instances/{linodeId}/backups/{backupId}/restore …with note on overwrite property about what happens if this is set to false

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3558,7 +3558,7 @@ paths:
                     If True, deletes all Disks and Configs on the target Linode
                     before restoring.
 
-                    If False, and the Disk image size is bigger than the available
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -3554,9 +3554,13 @@ paths:
                   example: 234
                 overwrite:
                   type: boolean
-                  description: >
+                  description: |
                     If True, deletes all Disks and Configs on the target Linode
                     before restoring.
+
+                    If False, and the Disk image size is less than the available
```
yoshizzle: shouldn't this be if the disk image size is MORE than the available space on the Linode?

PR #119, opened 2019-08-30
[Update] POST /linode/instances/{linodeId}/migrate

leslitagordita:
* Documentation for [ARB-1415](https://jira.linode.com/browse/ARB-1415)
* Also updates endpoint's description to add more information about pending host migration vs DC migration and info about NGN feature support.
* Updates endpoints summary name to include DC migration capability.
* Add request body example and linode cli example

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4941,13 +4941,25 @@ paths:
     x-linode-cli-command: linodes
     post:
       x-linode-grant: read_write
-      summary: Initiate Pending Migration
+      summary: Initiate Pending Host Migration/DC Migration
       description: >
-        In some circumstances, a Linode may have pending migrations scheduled that
-        you can initiate when convenient.  In these cases, a Notification
-        will be returned from [GET /account/notifications](#getNotifications).
-        This endpoint initiates the scheduled migration, which will shut the Linode
-        down, migrate it, and then bring it back to its original state.
+        Initiate a pending host migration that has been scheduled by Linode or
+        initiate a cross data center (DC) migration.  A list of pending migrations,
+        if any, can be accessed from [GET /account/notifications](/api/v4/account-notifications).
+        When the migration begins, your Linode will be shutdown if not already off.
+        If the migration initiated the shutdown, it will reboot the Linode when completed.
+
+
+        To initiate a cross DC migration, you must pass a `region` parameter to the
+        request body specifying the target data center region.
+        You can view a list of all available regions and their feature capabilities
+        from [GET /regions](/api/v4/regions). If your Linode has a pending migration already queued,
+        you will not be able to initiate a DC migration until it has completed.
+
+
+        **Note:** Next Generation Network (NGN) data centers do not support IPv6 `/116` pools or IP Failover.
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -4941,13 +4941,25 @@ paths:
...
+
+        To initiate a cross DC migration, you must pass a `region` parameter to the
+        request body specifying the target data center region.
+        You can view a list of all available regions and their feature capabilities
+        from [GET /regions](/api/v4/regions). If your Linode has a pending migration already queued,
+        you will not be able to initiate a DC migration until it has completed.
+
+
+        **Note:** Next Generation Network (NGN) data centers do not support IPv6 `/116` pools or IP Failover.
+        If you have these features enabled on your Linode and attempt to migrate to an NGN data center,
```
yoshizzle: @alsaihn can you confirm this? I thought we just wouldn't persist the IPv6 pool and IP failover, not that we would prevent the migration from happening. I did a quick check in the code and didn't see anything that would fail the migration.

leslitagordita: @yoshizzle I found that information in this related [KB post](https://kb.linode.com/pages/viewpage.action?pageId=42929211). I also checked w/ Sal, but more verification makes sense. Thanks!

alsaihn: There's a ticket for that exact behavior in our backlog, so the comment as listed here is accurate for now. It will have to change later though.

openapi.yaml
```diff
@@ -4969,6 +4981,8 @@ paths:
                     by using the [GET /regions](/api/v4/regions) endpoint.
                     A cross-region migration will cancel a pending migration
                     that has not yet been initiated.
+
+                    A cross-region migration will initiate a `linode_migrate_datacenter_create` event.
```
alsaihn: We're using "cross data center" instead of cross-region above. Either this or above should change to be consistent.

openapi.yaml
```diff
@@ -4941,13 +4941,25 @@ paths:
...
+        initiate a cross data center (DC) migration.  A list of pending migrations,
+        if any, can be accessed from [GET /account/notifications](/api/v4/account-notifications).
+        When the migration begins, your Linode will be shutdown if not already off.
+        If the migration initiated the shutdown, it will reboot the Linode when completed.
+
+
+        To initiate a cross DC migration, you must pass a `region` parameter to the
+        request body specifying the target data center region.
+        You can view a list of all available regions and their feature capabilities
+        from [GET /regions](/api/v4/regions). If your Linode has a pending migration already queued,
```
alsaihn: This may confuse readers since line 4946 says you can make changes if you have a pending migration. Better wording might be "If your Linode has a DC migration already queued, or you have initiated a previously scheduled migration, you will not be able to initiate additional migrations until it has completed."

openapi.yaml
leslitagordita: Good catch. Updated!

openapi.yaml
leslitagordita: That's better wording. Updated!

PR #121, opened 2019-09-04
new: Added documentation for the object-storage/buckets endpoints

Dorthu:
This is a draft; please review carefully and do not merge without
populating placeholder links.

These endpoints have existed for some time, but are finally being
documented (the last two are shipping on Monday).  I _do not_ advise
actual users to use these endpoints; they're mostly for the benefit for
cloud.  In keeping with API philosophy, they _are_ available to all
customers, but the s3 API will almost certainly be more useful in all
cases, which is why using these endpoints is discouraged.

The Linode CLI already supports all object storage operations through
the s3 API, so all of these endpoints are exempt from CLI builds.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -9034,7 +9034,8 @@ paths:
       x-code-samples:
       - lang: Shell
         source: >
-          curl -H "Authorization: Bearer $TOKEN" \
+          curl -H "Content-Type: application/json" \
+          -H "Authorization: Bearer $TOKEN" \
             -X POST -d '{
                 "cors_enabled": true,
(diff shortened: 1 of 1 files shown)


leslitagordita:
@Dorthu I'd like to push up some edits, but got a `permission denied` error. Would you mind adding me as a collaborator when you get a chance?

leslitagordita:
@Dorthu We add a few commits and wanted to confirm you've had a chance to review them (can't add a PR author as a reviewer) before hitting _merge_. Thank you!

Dorthu:
lgtm 👍

PR #123, opened 2019-09-06
Release/4.4.0

leslitagordita:
* Includes API docs ref release 4.4.0. Should not be merged until APInext v55.0 is released on or around 09/09


openapi.yaml
```diff
@@ -8862,9 +9410,10 @@ paths:
       servers:
       - url: https://api.linode.com/v4beta
       summary: Create an Object Storage Key
-      description: >
+      description: |
```
MichaelSaah: what do `>` and `|` mean here?

hzoppetti: The | respects line breaks without having to use markup where the > does not.

hzoppetti: This is actually not accurate, it's a yaml scalar block thing: https://yaml-multiline.info

PR #127, opened 2019-09-19
[Update] GET /object-storage/buckets/{clusterId}/{bucket}/object-list

leslitagordita:
* Add `next_marker` and `is_truncated` properties.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -9151,8 +9151,9 @@ paths:
         required: false
         description: >
           The "marker" for this request, which can be used to paginate through large buckets.
-          This should be the name of the last object on the last page of results. Listing
-          bucket contents *does not* support arbitrary page access.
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -9151,8 +9151,9 @@ paths:
         required: false
         description: >
           The "marker" for this request, which can be used to paginate through large buckets.
-          This should be the name of the last object on the last page of results. Listing
-          bucket contents *does not* support arbitrary page access.
+          This should be the name of the last object in the last page of results. Listing
```
Dorthu: This should actually be the `next_marker` returned with the last page; these original instructions are only _usually_ accurate.

leslitagordita: Thanks @Dorthu. Updated for your review.

PR #131, opened 2019-09-24
allOf Fixes

hzoppetti:


openapi.yaml
```diff
@@ -503,12 +503,12 @@ paths:
               allOf:
               - $ref: '#/components/schemas/CreditCard'
               - type: object
-              properties:
-                cvv:
-                  type: string
-                  description: >
-                    The Card Verification Value on the back of the card.
-                  example: '123'
+                properties:
```
nmelehan: hey heather, I think we need to keep the lines below the `- type: object` line in-line with that array item. For example, this is what it looks like in JSON with the indent i inserted:

```{
   "content": {
      "application/json": {
         "schema": {
            "allOf": [
               {
                  "$ref": "#/components/schemas/CreditCard"
               },
               {
                  "type": "object",
                  "properties": {
                     "cvv": {
                        "type": "string",
                        "description": "The Card Verification Value on the back of the card.\n",
                        "example": "123"
                     }
                  }
               }
            ]
         }
      }
   }
}
```

vs. without: 

```{
   "content": {
      "application/json": {
         "schema": {
            "allOf": [
               {
                  "$ref": "#/components/schemas/CreditCard"
               },
               {
                  "type": "object"
               }
            ],
            "properties": {
               "cvv": {
                  "type": "string",
                  "description": "The Card Verification Value on the back of the card.\n",
                  "example": "123"
               }
            }
         }
      }
   }
}
```

In the without indentation scenario, `properties` isn't in the same dictionary with the previous `type:object`

I've done this indenting a few times in this commit

openapi.yaml
```diff
@@ -2416,7 +2416,6 @@ paths:
               - type
               allOf:
               - $ref: '#/components/schemas/DomainRecord'
-              - type: object
```
nmelehan: I wasn't quite sure what was being gained by adding these `type: objects` in allOfs with only one other $ref. It seems like it's kind of like telling the file to do a union of a set of things (the $ref) and an empty set (`type: object`)--which would just be the set of things?

openapi.yaml
```diff
@@ -8371,7 +8367,7 @@ paths:
                           items:
                             type: object
                             allOf:
-                            $ref: '#/components/schemas/NodeBalancerNode'
+                            - $ref: '#/components/schemas/NodeBalancerNode'
```
nmelehan: allOf should accept an array, so I think we needed a dash here

openapi.yaml
```diff
@@ -9423,11 +9418,12 @@ paths:
           content:
             application/json:
               schema:
-                type: object
-                properties:
-                  secret_key:
-                    type: string
-                    example: OiA6F5r0niLs3QA2stbyq7mY5VCV7KqOzcmitmHw
+                allOf:
```
nmelehan: it seems like we do need to have the full schema for an object storage key here. In either case, the response isn't showing the contents of the schema, but that seems like a front-end bug?

openapi.yaml
hzoppetti: Sounds ok to me

openapi.yaml
hzoppetti: Yeah - I think Alban is going t work on this. I'm unable to get this response to show. There are other cases of this as well.

openapi.yaml
hzoppetti: ok

openapi.yaml
hzoppetti: We were only doing it for consistency? If you don't think we need it, then they can be removed. I'm for either case.

PR #133, opened 2019-09-26
[New] GET /managed/stats

hzoppetti:
Added GET /managed/stats endpoint.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -7374,6 +7374,83 @@ paths:
       - lang: CLI
         source: >
           linode-cli managed service-enable 9994
+  /managed/stats:
+    description: |
+      A collection of managed stats.
+    x-linode-cli-command: managed
+    get:
+      x-linode-grant: unrestricted only
+      tags:
+      - Managed
+      summary: List Managed Stats
+      description: |
+        Returns a list of Managed Stats on your Account. These
+        stats reflect the last 24 hours of combined usage for the following data:
+
+
+        * cpu
+        * disk
+        * swap
+        * network in
+        * network out
+      operationId: getManagedStats
+      x-linode-cli-action: stats-list
+      security:
+      - personalAccessToken: []
+      - oauth:
+        - account:read_only
+      responses:
+        '200':
+          description: A list of Managed Stats from the last 24 hours.
+          content:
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -7374,6 +7374,58 @@ paths:
...
+        - account:read_only
+      responses:
+        '200':
+          description: A list of Managed Stats from the last 24 hours.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  stats_data:
```
andystevensname: I think this is just `data`, rather than the stats_data object.

openapi.yaml
```diff
@@ -7374,6 +7374,58 @@ paths:
...
+        * network out
+      operationId: getManagedStats
+      x-linode-cli-action: stats-list
+      security:
+      - personalAccessToken: []
+      - oauth:
+        - account:read_only
+      responses:
+        '200':
+          description: A list of Managed Stats from the last 24 hours.
```
andystevensname: Can we describe that these are graph points here?

openapi.yaml
```diff
@@ -7374,6 +7374,58 @@ paths:
...
+    description: |
+      A collection of managed stats.
+    x-linode-cli-command: managed
+    get:
+      x-linode-grant: unrestricted only
+      tags:
+      - Managed
+      summary: List Managed Stats
+      description: |
+        Returns a list of Managed Stats on your Account in the form of y and x graph data points.
```
yoshizzle: probably best to either explain what the x and y points are (and put them in x, y order instead of y, x cause it's confusing that way), or just say something like:

Returns a list of Managed Stats on your Account. These stats reflect the last 24...

Either way works for me.

openapi.yaml
```diff
@@ -16461,6 +16513,61 @@ components:
...
+          readOnly: true
+          description: >
+            A stats graph data point.
+            example: 29.94
+        x:
+          type: integer
+          readOnly: true
+          description: >
+            A stats graph data point.
+          example: 11513761600000
```
yoshizzle: I would switch the order of these. People naturally want to see x before y, regardless of how they are returned in the api.

PR #140, opened 2019-10-04
Release 4.6.0

andystevensname:


openapi.yaml
```diff
@@ -4996,7 +4956,9 @@ paths:
 
         **Note:** Next Generation Network (NGN) data centers do not support IPv6 `/116` pools or IP Failover.
         If you have these features enabled on your Linode and attempt to migrate to an NGN data center,
-        the migration will not initiate. NGN data centers include Toronto and Mumbai.
+        the migration will not initiate. NGN data centers include Toronto and Mumbai. If a Linode cannot be
+        migrated because of an incompatibility, the user will be prompted to select a different data center
```
yoshizzle: I would change "the user" to "you"

Also, I'm not sure I have a great answer, but I'm not sure the specific listing of Mumbai and Toronto is a good solution. First, Newark is also partially NGN, and second this is going to have to remember to be changed when we add new NGNs (Sydney, etc.)

PR #146, opened 2019-10-21
Add upgrade field to POST linode/instances/{linodeId}/migrate

nmelehan:


openapi.yaml
```diff
@@ -5028,6 +5028,22 @@ paths:
...
                   example: us-east
+                upgrade:
+                  type: boolean
+                  description: >
+                    When initiating a cross DC migration, setting this value to
+                    true will also ensure that the Linode is upgraded to the latest
+                    generation of hardware that corresponds to your Linode's Type, if
+                    any free upgrades are available for it.
+
+                    If no free upgrades are available, and this value is set to true,
```
alsaihn: This changed during development to be consistent with the existing upgrade path. If no free upgrades are available then the code will return a 400 stating "No upgrades available" and not complete the migration.

nmelehan: thank you! addressed in f048179

PR #150, opened 2019-10-28
[Update] GET /account/settings

hzoppetti:
added object storage to GET /account/settings

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -12271,6 +12271,13 @@ components:
             or later.
           example: true
           x-linode-cli-display: 4
+        object_storage:
+          type: string
+          description: >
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -12273,13 +12273,14 @@ components:
           x-linode-cli-display: 4
         object_storage:
           type: string
+          readOnly: true
```
nmelehan: @Dorthu I added readOnly to this--is that accurate? when I read the code it didn't seem like it would accept changes to this value in PUT /account/settings

Dorthu: This is accurate - you enable object storage by using it, and you disable it by POSTing to object-storage/cancel

PR #158, opened 2019-11-14
Update GET /networking/ipv6/pools

hzoppetti:
added route_target to ipv6 pool and clarified ipv6 language in
GET /networking/ipv6/pools and
GET /networking/ipv6/ranges

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -7821,7 +7821,7 @@ paths:
       summary: List IPv6 Pools
       description: >
         Displays the IPv6 pools on your Account. A pool of IPv6 addresses are routed to all of your
-        Linodes in a single Region. Any Linode on your Account may bring up any address in this
+        Linodes in a single [Region](https://developers.linode.com/api/v4/regions). Any Linode on your Account may bring up any address in this
         pool at any time, with no external configuration required.
       operationId: getIPv6Pools
       x-linode-cli-action: v6-pools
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -7820,7 +7820,9 @@ paths:
       - Networking
       summary: List IPv6 Pools
       description: >
-        Displays the IPv6 pools on your Account.
+        Displays the IPv6 pools on your Account. A pool of IPv6 addresses are routed to all of your
+        Linodes in a single Region. Any Linode on your Account may bring up any address in this
```
yoshizzle: NB: I could see a benefit of linking the word Region to the LIST Regions endpoint.

openapi.yaml
```diff
@@ -7865,7 +7867,9 @@ paths:
       - Networking
       summary: List IPv6 Ranges
       description: >
-        Displays the IPv6 ranges on your Account.
+        Displays the IPv6 ranges on your Account. A range of IPv6 addresses is routed to a single
+        Linode in a given Region. Your Linode is responsible for routing individual addresses in
```
yoshizzle: Same as above.

hzoppetti: Added!

openapi.yaml
hzoppetti: Added!

PR #160, opened 2019-11-19
Cleaned up LKE Cluster Create endpoint

Dorthu:
The CLI didn't like how the spec was formatted initially, and I'm pretty
sure it was wrong.  This works with the linked CLI PR, but please verify
it looks good in the docs - I'm happy to work with someone to get this
into a mergable state.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -6050,6 +6050,11 @@ paths:
               - node_pools
               allOf:
               - $ref: '#/components/schemas/LKEClusterRequestBody'
+              - properties:
+                  node_pools:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/LKENodePoolRequestBody'
       responses:
         '200':
           description: Kubernetes cluster creation has started.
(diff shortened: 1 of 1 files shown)


leslitagordita:
* Thanks for this update, Will. `node_pools` is certainly a list of objects. That was an error on my part.
* I originally had `node_pools` as a property directly in the `POST /lke/clusters request body`, and not in the `LKEClusterRequestBody`. If`node_pools` is in the `LKEClusterRequestBody` then `node_pools` is also displayed in the response for other endpoints: 
  * POST /lke/clusters
  * GET /lke/clusters
  * GET /lke/clusters/{clusterId}
  * PUT /lke/clusters/{clusterId}
* If you think it is more correct to have all properties in a component schema object, I can create a third schema object for all LKE cluster's common properties and convert LKECluster to LKEClusterResponse OR
* Move `node_pools` back to `POST /lke/clusters request body`, but of course keep your update to its datatype. 
* OR is there a third thing I'm completely not considering?

Dorthu:
Let me move it back to where it was originally and see if the CLI will still use it - that's probably the right place for it.

Dorthu:
@leslitagordita fixed the placement of `node_pools` - cli still likes it.

PR #164, opened 2019-11-27
[Update] LKENodePool to use multi-item example

leslitagordita:
This example better represents the `linodes` property which returns an array of objects.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -15660,6 +15660,13 @@ components:
             its ID will be returned as null.
           items:
             $ref: '#/components/schemas/LKELinodeStatus'
+          example:
+            - id: 123456
+              status: ready
+            - id : 123457
+              status: ready
+            - id: 123458
+              status: not_ready
     LKELinodeStatus:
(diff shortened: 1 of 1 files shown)


leslitagordita:
*Note:* This requires a front end fix before it can be merged in.

displague:
A `linode_id: null` example would also be useful to see.

alioso:
hi @leslitagordita ! I found out I am still assigned this review :) can we please close this?

alioso (approved):

asauber (changes requested):
Looks like this should be closed.

displague:
Yes, this PR offers an example for the `linodes` field in the `LKENodePool` schema.
The example is now present through the `LKENodeStatus` schema property examples.  
This schema is embedded in the `nodes` property which replaced the `linodes` property in the OAS3 definition.

displague:
However, if it is still possible for an `id` to be null, that is a case that is not well documented in the spec or the examples.

PR #170, opened 2019-12-06
Remove null from enum values

phillc:
Co-Authored-By: William Smith <dorthu@users.noreply.github.com>

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -14021,7 +14021,6 @@ components:
           type: string
           nullable: true
           enum:
-          - null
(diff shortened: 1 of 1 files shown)


Dorthu:
For clarity, this was breaking the `openapi-generator-cli` when attempting to generate a ruby library from our spec.  While the standard isn't clear about it, I suspect that defining an enum as `nullable: true` is enough to suggest that one of its values may be `null`, and you don't need to explicitly redefine that in the `enum:` block.

PR #171, opened 2019-12-13
[Update] Get/Put Account - Add euuid

hzoppetti:
added euuid to account

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -12995,7 +12995,7 @@ components:
         euuid:
           type: string
           description: >
-            An external unique identifier for this customer.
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -12992,6 +12992,13 @@ components:
             If you do not live in a country that collects tax, this should be `null`.
           maxLength: 100
           example: ATU99999999
+        euuid:
+          type: string
+          description: >
+            An external unique identifier for this customer.
```
alsaihn: Should we use `Account` here instead of `customer`?

hzoppetti: done

PR #177, opened 2019-12-18
[Update] last_successful return None to null when none found

hzoppetti:
changed None to null for last successful when no last successful backup date is found

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -15129,7 +15129,7 @@ components:
             last_successful:
               type: string
               format: date-time
-              description: The last successful backup date. 'None' if there was no previous backup.
+              description: The last successful backup date. 'null' if there was no previous backup.
               readOnly: true
               example: '2018-01-01T00:01:01'
         watchdog_enabled:


openapi.yaml
```diff
@@ -15129,7 +15129,7 @@ components:
             last_successful:
               type: string
```
nmelehan: hey @hzoppetti, I'm wondering what your opinion on adding `nullable: true` to this property might be. I'm a little undecided about it, because my reading of the `nullable` documentation from Swagger is that it might be used more on non-readonly properties; it would communicate that the API user can send `null` for the property:

https://swagger.io/docs/specification/data-models/data-types/#null
https://swagger.io/specification/#schemaNullable

hzoppetti: right - it is technically nullable - but only by the system, not by users...

The openapi spec says "Allows sending a null value for the defined schema. Default value is false." so I vote no for now.

nmelehan: thank you! will merge as-is

PR #184, opened 2020-01-09
[Update] Invoice Items

hzoppetti:
In InvoiceItem:
- updated integer to number for amount, tax, total, and unit price 
- updated examples to reflect number type
- changed type to exclude prepay

In Invoice:
- updated integer to number for total, tax, and subtotal
- updated examples to reflect number type


nmelehan:
@cschreep should be ready for review

PR #185, opened 2020-01-09
[Update] Domain status enum

hzoppetti:
updated domain status enum to only include active and disabled

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -13501,8 +13501,6 @@ components:
           enum:
           - disabled
           - active
-          - edit_mode
-          - has_errors
           description: >
             Used to control whether this Domain is currently being rendered.
           example: active


nmelehan:
@cschreep Hola! This should be ready for your review, though I do have one question for you--

Technically, it looks like the API code updates only get rid of setting the `edit_mode` status on the API and doesn't have any bearing on whether a domain with status of `edit_mode` could be read from the API, if that were to exist. Would you prefer it if we set `edit_mode` to read-only instead of getting rid of it? Or, because we're looking to phase it out in general, should we just stick to getting rid of it entirely?

cschreep:
@nmelehan the request for this change suggests that we don't want it at all anymore, so we should stick with getting rid of it entirely

PR #192, opened 2020-01-16
[Update] Oauth section update

hzoppetti:
Updated the OAuth section to include:
- link to docs
- flow for private vs public
- before you begin info on creating a client
- info on refresh tokens
- links to endpoints


openapi.yaml
```diff
@@ -47,33 +47,45 @@ info:
     | **HTTP Authorization Scheme** | bearer |
 
     ## OAuth
+    If all you need is a Personal Access Token, feel free to skip ahead to the next section.
 
-    The OAuth workflow is a three-step process to authenticate a User before an
-    application can start making API calls on the User's behalf. If all you need
-    is a Personal Access Token, feel free to skip ahead to the next section.
+    For more help on OAuth see our guide on [How to Create an OAuth App with teh Linode Python API Library](https://www.linode.com/docs/platform/api/how-to-create-an-oauth-app-with-the-linode-python-api-library/#oauth-2-authentication-exchange).
```
Dorthu: > with teh Linode

openapi.yaml
```diff
@@ -47,33 +47,45 @@ info:
...
 
-    First, the User visits the application's website and is directed to log with
-    Linode. The User is then redirected to Linode's authentication server and
-    presented the scope levels the application is requesting. Once the User
-    accepts the request for access, we redirect them back to the application's
-    specified redirect URI with an access code.
+    Before you begin, you need to [Create an OAuth Client](https://developers.linode.com/api/v4/account-oauth-clients/#post).
```
Dorthu: Might also consider https://cloud.linode.com/profile/clients

openapi.yaml
```diff
@@ -47,33 +47,45 @@ info:
...
+    | 3.  The user logs into the authorization server with their username and password. | 3.  The user logs into the authorization server with their username and password. |
+    | 4.  The authorization server redirects the user to the specificed redirect URL with a temporary authorization `code` (exchange code) in the URL. | 4.  The authorization server redirects the user back to your applaction with an OAuth token in the URL hash. This is temporary and expires in 2 hours without a `refresh_token`. |
+    | 5.  The application issues a POST request (*see below*) to the authentication server with the exchange code, `client_id`, and the client application's `client_secret`. In return, the server returns an OAuth token and `refresh_token`. |  |
+    | 6.  The authentication server responds to the client application with a new OAuth `access_token`. `access_token` is set to expire in 2 hours if not refreshed. |  |
```
Dorthu: Actually, the token will just expire in 2 hours.  If refreshed, a new token is generated.

openapi.yaml
```diff
@@ -47,33 +47,45 @@ info:
...
+    | 7.  The refresh token can be used to get a new OAuth `access_token` for another 2 hours by contacting the authentication server again with the `client_id`, `client_secret`, and `refresh_token`. | |
```
Dorthu: You also get back a new refresh token, allowing you to continue to keep the session alive.

openapi.yaml
```diff
@@ -47,33 +47,45 @@ info:
...
+
+      Once the User has logged into Linode and you have received an exchange code,
```
Dorthu: Might want to qualify that this only applies to the private workflow.

openapi.yaml
```diff
@@ -47,33 +47,45 @@ info:
...
-    | PARAMETER | DESCRIPTION |
-    |-----------|-------------|
-    | client_id | Your app's client ID |
-    | client_secret | Your app's client secret |
-    | code | The code you just received from the redirect |
+      | PARAMETER | DESCRIPTION | FIND OR UPDATE THIS INFORMATION |
+      |-----------|-------------|---------------------------------|
+      | client_id | Your app's client ID. | [List your OAuth Clients](https://developers.linode.com/api/v4/account-oauth-clients). |
+      | client_secret | Your app's client secret. | [View your OAuth Client](https://developers.linode.com/api/v4/account-oauth-clients-client-id). [Reset Your OAuth Client Secret](https://developers.linode.com/api/v4/account-oauth-clients-client-id-reset-secret/#post). |
+      | code | The code you just received from the redirect | [View your OAuth Client](https://developers.linode.com/api/v4/account-oauth-clients-client-id). [Reset Your OAuth Client Secret](https://developers.linode.com/api/v4/account-oauth-clients-client-id-reset-secret/#post). |
```
Dorthu: These two links don't apply to the exchange code, and there's no way to retrieve it if you don't capture it in the redirect.

openapi.yaml
hzoppetti: fixed

openapi.yaml
hzoppetti: changed to include both options

openapi.yaml
hzoppetti: updated this to be more clear

openapi.yaml
hzoppetti: updated to be more clear

openapi.yaml
hzoppetti: added a note to say this extends the private workflow

openapi.yaml
hzoppetti: removed the link column because it was confusing

openapi.yaml
```diff
@@ -47,19 +47,43 @@ info:
...
+    | 4.  The authorization server redirects the user to the specificed redirect URL with a temporary authorization `code` (exchange code) in the URL. | 4.  The authorization server redirects the user back to your application with an OAuth `access_token` embedded in the redirect URL's hash. This is temporary and expires in 2 hours without a `refresh_token`. |
+    | 5.  The application issues a POST request (*see below*) to the authentication server with the exchange code, `client_id`, and the client application's `client_secret`. | |
+    | 6.  The authentication server responds to the client application with a new OAuth `access_token` and `refresh_token`. The `access_token` is set to expire in 2 hours. | |
+    | 7.  The `refresh_token` can be used to get a new OAuth `access_token`, good for another 2 hours, and a new `refresh_token`, which you can use to extend the session again by this same method, by contacting the authentication server again with the `client_id`, `client_secret`, and `refresh_token`. | |
```
nmelehan: it seems like we might be using `authentication server` and `authorization server` somewhat interchangeably here, which could maybe become confusing? I've read that OAuth makes distinctions between authentication and authorization:
https://stackoverflow.com/questions/33702826/oauth-authorization-vs-authentication

Maybe we can decide to just use one of these? I'm not so great with OAuth, so if a @Dorthu or @jnschaeffer could offer some insight too, that'd be great

nmelehan: I think step 7 is a bit confusing. I thought about splitting it into two sentences, but I'm not actually sure about what a refresh token is capable of doing. As written, it sounds like a refresh token can refresh itself... which doesn't seem right... but maybe it is?

openapi.yaml
```diff
@@ -47,19 +47,43 @@ info:
...
+    | 3.  The user logs into the authorization server with their username and password. | 3.  The user logs into the authorization server with their username and password. |
+    | 4.  The authorization server redirects the user to the specificed redirect URL with a temporary authorization `code` (exchange code) in the URL. | 4.  The authorization server redirects the user back to your application with an OAuth `access_token` embedded in the redirect URL's hash. This is temporary and expires in 2 hours without a `refresh_token`. |
+    | 5.  The application issues a POST request (*see below*) to the authentication server with the exchange code, `client_id`, and the client application's `client_secret`. | |
+    | 6.  The authentication server responds to the client application with a new OAuth `access_token` and `refresh_token`. The `access_token` is set to expire in 2 hours. | |
```
nmelehan: Just want to confirm with @dorthu or @jnschaeffer that it's correct that the token endpoint will return both an access token and a refresh token for private/confidential oauth clients (described in step 6 of the private workflow)

openapi.yaml
```diff
@@ -47,19 +47,43 @@ info:
...
+    | 1.  The User visits the application's website and is directed to login with Linode. | 1.  The User visits the application's website and is directed to login with Linode. |
+    | 2.  Your application then redirects the user to Linode's [authentication server](https://login.linode.com) with the client application's `client_id` and requested OAuth `scope`, which should appear in the URL of the login page. | 2.  Your application then redirects the user to Linode's [authentication server](https://login.linode.com) with the client application's `client_id` and requested OAuth `scope`, which should appear in the URL of the login page. |
+    | 3.  The user logs into the authorization server with their username and password. | 3.  The user logs into the authorization server with their username and password. |
+    | 4.  The authorization server redirects the user to the specificed redirect URL with a temporary authorization `code` (exchange code) in the URL. | 4.  The authorization server redirects the user back to your application with an OAuth `access_token` embedded in the redirect URL's hash. This is temporary and expires in 2 hours without a `refresh_token`. |
```
nmelehan: Can we more explicitly state that a refresh token is *not* returned by the login server (step 4 of the public workflow)

openapi.yaml
hzoppetti: I agree - this is confusing, our doc also uses both of these terms in this way, if they are different that's fine. If this is all the same, authentication server, we should update the doc as well.

openapi.yaml
hzoppetti: updated this to be more explicit

openapi.yaml
hzoppetti: Updated this to be more clear.

nmelehan: @hzoppetti and I talked to @jnschaeffer, resolved this in subsequent commits

PR #199, opened 2020-02-07
[Update] LKE node pools - nodes

hzoppetti:
changed linodes to nodes in node pools return


yoshizzle:
I don't think I'm the best person to ask for a review on this. I think Sauber and maybe @adammohammed or @jnschaeffer would be more appropriate.

PR #212, opened 2020-03-04
Add cURL and CLI examples to POST /tags

nmelehan:
Tagging with do-not-merge: this has some issues I couldn't figure out, and am asking for some second opinions

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -12036,8 +12036,10 @@ paths:
           source: >
             linode-cli tags create \
               --label 'example tag' \
-              --linodes (123 456) \
-              --volumes (9082 10003)
+              --linodes 123 \
+              --linodes 456 \
+              --volumes 9082 \
+              --volumes 10003
   /tags/{label}:
     x-linode-cli-command: tags
     parameters:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -12021,6 +12021,23 @@ paths:
                 $ref: '#/components/schemas/Tag'
         default:
           $ref: '#/components/responses/ErrorResponse'
+      x-code-samples:
+        - lang: Shell
+          source: >
+            curl -H "Content-Type: application/json" \
+                -H "Authorization: Bearer $TOKEN" \
+                -X POST -d '{
+                    "label": "example tag",
+                    "linodes": [123,456],
+                    "volumes": [9082,10003]
+                }' \
+                https://api.linode.com/v4/tags
+        - lang: CLI
+          source: >
+            linode-cli tags create \
+              --label 'example tag' \
+              --linodes (123 456) \
+              --volumes (9082 10003)
   /tags/{label}:
     x-linode-cli-command: tags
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -12021,6 +12021,23 @@ paths:
...
         default:
           $ref: '#/components/responses/ErrorResponse'
+      x-code-samples:
+        - lang: Shell
+          source: >
+            curl -H "Content-Type: application/json" \
+                -H "Authorization: Bearer $TOKEN" \
+                -X POST -d '{
+                    "label": "example tag",
+                    "linodes": [123,456],
```
nmelehan: @leslitagordita @yoshizzle I did some testing for this but ran into some problems:

When running the example command as it is here (with the IDs updated for resources that are actually on my account), the tag was created, but the Linodes and Volumes didn't get the tag applied to them. No error was returned by the API, and my API token had full permissions to write to my account. 

Can I ask you to try this out too? I'm not sure if this is a problem with the example I've added, or if it might be a bug in the API.

openapi.yaml
```diff
@@ -12021,6 +12021,23 @@ paths:
...
+                    "volumes": [9082,10003]
+                }' \
+                https://api.linode.com/v4/tags
+        - lang: CLI
+          source: >
+            linode-cli tags create \
+              --label 'example tag' \
+              --linodes (123 456) \
```
nmelehan: @Dorthu I realized when trying to write this example that I don't know how to represent lists as arguments for the Linode CLI. Using the Bash array syntax here didn't work; I got the following error: 

```
linode-cli tags create \
  --label 'example tag' \
  --linodes (123 456) \
  --volumes (9082 10003)
zsh: unknown file attribute: 1
```

Any tips for where I might be going wrong?

Dorthu: You need to specify `--linodes` multiple times; in this case you want: 

```base
linode-cli tags create \
  --label 'example tag' \
  --linodes 123 --linodes 456) \
  --volumes 9082 --volumes 10003
```

openapi.yaml
Dorthu: I ran this with Linode and Volume IDs that exist on my account, and it worked as I expected.

openapi.yaml
```diff
@@ -12083,6 +12102,11 @@ paths:
                     $ref: '#/components/schemas/PaginationEnvelope/properties/results'
         default:
           $ref: '#/components/responses/ErrorResponse'
+      x-code-samples:
+      - lang: Shell
+        source: >
+          curl -H "Authorization: Bearer $TOKEN" \
+              https://api.linode.com/v4/tags/example%20tag
```
Dorthu: You could also put the entire URL in quotes to avoid having to escape the space:

```suggestion
              "https://api.linode.com/v4/tags/example tag"
```

leslitagordita: Good suggestion. Updated!

PR #216, opened 2020-03-06
[Update] `root_pass` field password requirements

leslitagordita:
* Include bullet about the `root_pass` field's new password requirements.


leslitagordita:
DM'ed Carl to please review. He does not come up as a possible reviewer.

PR #228, opened 2020-03-25
Allow CLI to accept k8s version when creating clusters

Dorthu:
This is for https://github.com/linode/linode-cli/pull/177

Right now, the "version" argument accepted when creating an LKE Cluster
translates in the CLI to `--version`, which the CLI interprets to mean
it should print its own version and exit.

For this reason, a new spec extension has been added,
`x-linode-cli-name`, which is allowed in schemas to define a new name
for the CLI argument for any POST/PUT attribute.  To keep things
consistent between the API and CLI, this should be used only in extreme
circumstances (such as noted above).


leslitagordita:
@Dorthu when I test this locally (building the linode-cli from master), I receive an error:
`linode-cli: error: unrecognized arguments: --kube-version 1.16`

Dorthu:
Yeah, sorry, the associated CLI PR was closed because I'd already fixed this issue a different way 🙃 I'll close this as well.

PR #237, opened 2020-04-08
Minor LKE Schema fixes to bring it in line with API

adammohammed:
There were two main changes in this PR
1. The `LKENodeStatus` had `id` as a number when it is a string. That id is the name of the node object as listed by the Kubernetes API server.
2. Removed the `x-linode-filterable` key from fields which are not filterable.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -15812,11 +15812,11 @@ components:
       readOnly: true
       properties:
         id:
-          type: number
+          type: string
           description: >
             The Node's ID.
           x-linode-filterable: true
-          example: 123456
+          example: "123456"
         instance_id:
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -15794,7 +15791,6 @@ components:
           type: integer
           description: >
             This Node Pool's unique ID.
-          x-linode-filterable: true
```
nmelehan, line 15797: @adammohammed I'm seeing that the node pool ID is filterable in the API--should we revert this line?

adammohammed: Yep, this was a mistake good catch. Gonna double check the others too

PR #241, opened 2020-04-09
[New ] GET /lke/clusters/{clusterId}/api-endpoints

leslitagordita:


openapi.yaml
```diff
@@ -6585,6 +6585,67 @@ paths:
...
+      - oauth:
+        - lke:read_only
+      tags:
+      - Linode Kubernetes Engine (LKE)
+      servers:
+      - url: https://api.linode.com/v4beta
+      summary: List Kubernetes API Endpoints
+      description: >
+        List the Kubernetes API server endpoints for this cluster. Please note that it often takes
+        2-5 minutes before the endpoint is ready after first [creating a new cluster](/api/v4/lke-clusters/#post).
```
adammohammed, line 6611: /v4beta

openapi.yaml
```diff
@@ -6585,6 +6585,67 @@ paths:
...
+        updates in the future. This notice will be removed when this endpoint is out of
+        beta. Sign up for the beta [here](https://welcome.linode.com/lkebeta/).
+      responses:
+        '200':
+          description: Returns the Kubernetes API server endpoints for this cluster.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
```
adammohammed, line 6625: In addition to renaming this endpoint, the return structure changed to be more in line with the rest of the API.
Now the response looks something like this
```
{
  "results": 1,
  "pages": 1,
  "page": 1,
  "data": [{"endpoint": "https://207.xxx.xxx.xxx:443"},{"endpoint": "https://207.xxx.xxx.xxx:6443"}],
}
```

openapi.yaml
nmelehan: If this is in regard to the `creating a new cluster` link--I think that the `v4` in this URL is not reflective of the v4 vs v4beta for the endpoint URL

adammohammed: ah it's just an anchor I see.

PR #253, opened 2020-04-21
[Update] Profile collection

leslitagordita:
* Add info about GitHub `authentication_type`
* This feature will be going live soon in Cloud Manager, so users will be able to receive `github` as a possible value in their Profile collection responses.


salazarleslie:
I would like to add a link to the docs library guide on enabling TPA. Going to hold off on merging to find out timing for when that doc will go live, so neither block each other from being released.

PR #259, opened 2020-05-15
[Update] Profile collection lish_auth_method field

leslitagordita:
* When third party authentication is enabled, you cannot use `password_keys` as your `lish_auth_method`.


leslitagordita:
@bbiggerr Please review :)

PR #274, opened 2020-06-25
Add minimum and maximum values for DomainRecord's weight

nmelehan:
One of the last sprint's tasks was to standardize the max weight of a DomainRecord to 65535. The minimum and maximum weight were missing from the spec file, so this PR fixes that too

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -13797,6 +13797,8 @@ components:
           description: >
             The relative weight of this Record. Higher values are preferred.
           example: 50
+          minimum: 0
+          maximum: 65535
           x-linode-cli-display: 7
         port:
           type: integer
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -13797,6 +13797,8 @@ components:
           description: >
             The relative weight of this Record. Higher values are preferred.
           example: 50
+          minimum: 0
+          maximum: 65535
           x-linode-cli-display: 7
         port:
           type: integer
```
alsaihn, line 13804: While we didn't change it as part of this ticket, the port min/max is the same as the weight min/max. Can we add that here?

nmelehan: sure, will do

nmelehan: @alsaihn added in 14cb3b7

PR #276, opened 2020-06-26
New: /account/logins endpoints

nmelehan:


openapi.yaml
```diff
@@ -877,6 +877,85 @@ paths:
...
+      schema:
+        type: integer
+    x-linode-cli-command: account
+    get:
+      tags:
+      - Account
+      summary: View Login
+      description: >
+        Returns a login object that displays information about a successful
+        login from an arbitrary user on the account. This command can only be
```
yoshizzle, line 934: absolute nitpick here. "Arbitrary" user sounds weird to me, almost implying that the user isn't directly tied to the login object identified by the id.

It's a non-breaking thought, so if you feel like it stands well as-is then I'm all for it.

nmelehan: I see what you mean... maybe an alternative could be:

Returns a Login object that displays information about a successful login. The logins that can be viewed can be for any user on the account, and are not restricted to only the logins of the user that is accessing this API endpoint. This command can only be accessed by the unrestricted users of the account.

yoshizzle: that works for me

PR #284, opened 2020-07-10
[New] Rate Limiting Section

hzoppetti:
initial check in

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -327,9 +327,6 @@ info:
     With the Linode API, you can make up to 1600 general API requests every two minutes per user as
     determined by IP adddress or by OAuth token.
 
-    If a user has multiple OAuth tokens, the quota of requests per user is shared amongst all tokens
-    for the user.
-
     **Note:** There may be rate limiting applied at other levels outside of the API, for example, at the load balancer.
 
     `/stats` endpoints have their own dedicated limits of 100 requests per minute per user.
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -330,6 +330,8 @@ info:
     If a user has multiple OAuth tokens, the quota of requests per user is shared amongst all tokens
     for the user.
 
+    **Note:** There may be rate limiting applied at other levels outside of the API, for example, at the load balancer.
+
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -322,6 +322,39 @@ info:
         }'
     ```
 
+    # Rate Limiting
+
+    With the Linode API, you can make up to 1600 general API requests every two minutes per user as
+    determined by IP adddress or by OAuth token.
+
+    If a user has multiple OAuth tokens, the quota of requests per user is shared amongst all tokens
+    for the user.
```
er0k, line 331: I don't think this is correct... our limiter `key_func` (located [here](https://bits.linode.com/LinodeAPI/apinext/blob/development/apinext/services/limiting/__init__.py#L17-L19)) takes an md5 hash of the bearer token and uses that to determine the "bucket in which each request is put into when evaluating whether it is within the rate limit or not" (from the [limiter docs](https://flask-limiter.readthedocs.io/en/stable/#rate-limit-domain)), so I think if a user has multiple OAuth token, the quota or requests per users is _unique_ for each token.

openapi.yaml
```diff
@@ -322,6 +322,36 @@ info:
...
+    **Note:** There may be rate limiting applied at other levels outside of the API, for example, at the load balancer.
+
+    `/stats` endpoints have their own dedicated limits of 100 requests per minute per user.
+    These endpoints are:
+
+    * [View Linode Statistics](https://api.linode.com/v4/linode/instances/{linodeId}/stats)
+    * [View Linode Statistics (year/month)](https://api.linode.com/v4/linode/instances/{linodeId}/stats/{year}/{month})
+    * [View NodeBalancer Statistics](https://api.linode.com/v4/nodebalancers/{nodeBalancerId}/stats)
+    * [List Managed Stats](https://api.linode.com/v4/managed/stats)
+
```
yoshizzle, line 339: support tickets POST also has a limit: https://bits.linode.com/LinodeAPI/apinext/blob/development/apinext/blueprints/support/tickets.py#L136

hzoppetti: Thank you - I added a section for support tickets

PR #289, opened 2020-07-13
[Update] lke/clusters PUT and lke/cluster POST to reflect actual behavior

sibucan:
LKE does not impose ASCII-only restrictions on the label name. The only restrictions on the label are length and JSON validity.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -16023,9 +16023,6 @@ components:
                 characters.
           minLength: 1
           maxLength: 32
-          # Kubernetes does NOT allow underscores in resource names and we are
-          # passing this name to Kubernetes. Thus the following pattern:
-          pattern: '[a-zA-Z0-9-]{3, 32}'
           x-linode-cli-display: 2
           example: lkecluster12345
         region:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -15933,10 +15933,18 @@ components:
 
             Labels have the following constraints:
 
-              * Must start with an alpha character
-              * Must only consist of alphanumeric characters and dashes (`-`)
-              * Must not contain two dashes in a row
-          minLength: 3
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -15933,14 +15933,12 @@ components:
 
             Labels have the following constraints:
 
-              * Must start with an alpha character
-              * Must only consist of alphanumeric characters and dashes (`-`)
-              * Must not contain two dashes in a row
-          minLength: 3
+              * 4-byte UTF-8 characters are not supported.
+              * UTF-8 characters will be encoded in their byte representation.
```
asauber, line 15937: This should be 
> UTF-8 characters will be returned by the API using escape sequences of their Unicode Code points.

You should then include an example of a 3 byte character that has a 2 byte code point to drive the example home.

openapi.yaml
```diff
@@ -15933,14 +15933,12 @@ components:
 
             Labels have the following constraints:
 
-              * Must start with an alpha character
-              * Must only consist of alphanumeric characters and dashes (`-`)
-              * Must not contain two dashes in a row
-          minLength: 3
+              * 4-byte UTF-8 characters are not supported.
+              * UTF-8 characters will be encoded in their byte representation. 
+              * If the label is using UTF-8, the response will return the value
```
asauber, line 15938: > The Unicode code point escape sequences returned by the API use up to 193 1 byte characters

Then give an example here of such a response (with less than 193 byte characters)

openapi.yaml
asauber: For example the Japanese character か is encoded with 3 bytes in UTF-8 (0xE382AB). It's Unicode code point is 2 bytes (0x30AB). We support this character and we will return it as the escape sequence using six 1 byte characters which represent 2 bytes of Unicode code point ("\u30ab")

PR #294, opened 2020-07-23
[NEW] POST /lke/clusters/{clusterId}/pools/{poolId}/recycle

hzoppetti:
new lke node pool recycle endpoint

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -6730,20 +6730,20 @@ paths:
     parameters:
     - name: clusterId
       in: path
-      description: ID of the Kubernetes cluster to this Node Pool is attached to.
+      description: ID of the Kubernetes cluster this Node Pool is attached to.
       required: true
       schema:
         type: integer
     - name: poolId
       in: path
-      description: ID of the Kubernetes pool to be recycled.
+      description: ID of the Node Pool to be recycled.
       required: true
       schema:
         type: integer
     x-linode-cli-command: lke
     post:
-      operationId: postLKEClusterPoolsRecycle
-      x-linode-cli-action: cluster-recycle
+      operationId: postLKEClusterPoolRecycle
+      x-linode-cli-action: pool-recycle
       security:
       - personalAccessToken: []
       - oauth:
@@ -6767,7 +6767,7 @@ paths:
                 https://api.linode.com/v4/lke/clusters/12345/pools/456/recycle
       - lang: CLI
         source: >
-          linode-cli lke cluster-recycle 12345 456
+          linode-cli lke pool-recycle 12345 456
   /lke/clusters/{clusterId}/api-endpoints:
     parameters:
     - name: clusterId
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -6751,8 +6751,12 @@ paths:
       tags:
       - Linode Kubernetes Engine (LKE)
       summary: Recycle Node Pool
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -6726,6 +6726,48 @@ paths:
       - lang: CLI
         source: >
           linode-cli lke pool-delete 12345 456
+  /lke/clusters/{clusterId}/pools/{poolId}/recycle:
+    parameters:
+    - name: clusterId
+      in: path
+      description: ID of the Kubernetes cluster to this Node Pool is attached to.
```
g1bl3t, line 6733: This reads a lil funny, I think the "to" can be removed, like
"ID of the Kubernetes cluster this Node Pool is attached to."

openapi.yaml
```diff
@@ -6726,6 +6726,48 @@ paths:
...
+      required: true
+      schema:
+        type: integer
+    - name: poolId
+      in: path
+      description: ID of the Kubernetes pool to be recycled.
```
g1bl3t, line 6739: Can we change this to "ID of the Node Pool to be recycled."? since Node Pools are a provider construct and not actually a K8s concept.

openapi.yaml
```diff
@@ -6726,6 +6726,48 @@ paths:
...
+      required: true
+      schema:
+        type: integer
+    x-linode-cli-command: lke
+    post:
+      operationId: postLKEClusterPoolsRecycle
```
g1bl3t, line 6745: s/Pools/Pool, since this endpoint can only recycle a single Node Pool

openapi.yaml
```diff
@@ -6726,6 +6726,48 @@ paths:
...
+      x-code-samples:
+      - lang: Shell
+        source: >
+          curl -H "Content-Type: application/json" \
+                -H "Authorization: Bearer $TOKEN" \
+                -X POST \
+                https://api.linode.com/v4/lke/clusters/12345/pools/456/recycle
+      - lang: CLI
+        source: >
+          linode-cli lke cluster-recycle 12345 456
```
g1bl3t, line 6770: is "cluster-recycle" set in stone already? I think I like "pool-recycle", since this endpoint only recycles a single Node Pool

openapi.yaml
```diff
@@ -6726,6 +6726,48 @@ paths:
...
+    - name: poolId
+      in: path
+      description: ID of the Kubernetes pool to be recycled.
+      required: true
+      schema:
+        type: integer
+    x-linode-cli-command: lke
+    post:
+      operationId: postLKEClusterPoolsRecycle
+      x-linode-cli-action: cluster-recycle
```
g1bl3t, line 6746: same as below, can we change this to "pool-recycle"? might be a question for the CLI team

openapi.yaml
hzoppetti: fixed

openapi.yaml
hzoppetti: fixed

openapi.yaml
hzoppetti: fixed

openapi.yaml
hzoppetti: updated

openapi.yaml
hzoppetti: updated

PR #316, opened 2020-08-31
Filled in IPv6Range object

Dorthu:
Closes https://github.com/linode/linode-cli/issues/207

It was reported above that the CLI wasn't showing complete output for
IPv6 Ranges compared to IPv6 Pools.  Turns out this is because the docs
were missing several fields for ranges.

This PR adds the missing fields to IPv6Ranges.  I based the descriptions
and values off of the equivalent fields in IPv6Pool.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -15264,7 +15264,7 @@ components:
           type: string
           description: >
             The IPv6 range of addresses in this pool.
-          example: '2600:3c01::02:5000::'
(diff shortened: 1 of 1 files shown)


leslitagordita (approved):
@Dorthu added some clarification to the List IPv6 Ranges endpoint re: `/64` block for an IPv6 range and the requirement to open up a support ticket.

Dorthu:
Thank you! 👍

PR #317, opened 2020-08-31
[Update] NodeBalancer Config with ProxyProtocol

hzoppetti:
updated NodeBalancer config to include ProxyProtocol

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -17146,12 +17146,24 @@ components:
           example: true
           x-linode-cli-display: 6
         proxy_protocol:
+          description: >
+            ProxyProtocol is an internet protocol that sends information as a header on a
+            TCP connection without losing client data. When used with a NodeBalancer, it
+            enables the termination of TCP connections.
+
+
+            Linode NodeBalancers supports two versions, `v1` and `v2`. `v1` is human readable
+            text while `v2` makes use of a binary signature.
+
+
+            If ommited from the configuration, the default is `none`.
           type: string
           enum:
           - v1
           - v2
           - none
-          example: v1
+          example: none
+          default: none
         cipher_suite:
           type: string
           enum:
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -17139,6 +17145,13 @@ components:
             enough for it to be considered unhealthy and taken out of rotation.
           example: true
           x-linode-cli-display: 6
+        proxy_protocol:
```
aspsk, line 17148: @hzoppetti it accepts `"v1"`, `"v2"`, and `"none"`, so it is not nullable

hzoppetti: updated enum

openapi.yaml
```diff
@@ -17139,6 +17145,25 @@ components:
...
+          description: >
+            ProxyProtocol is an internet protocol that sends information as a header on a
+            TCP connection without losing client data. When used with a NodeBalancer, it
+            enables the termination of TCP connections.
+
+
+            Linode NodeBalancers supports two versions, `v1` and `v2`. `v1` is human readable
```
aspsk, line 17155: Another comment. Comparing this to other descriptions in the same section (e.g. `stickiness`) , this is better be formatted as

 *  If set to `none` NodeBalancer doesn't send any auxiliary data over TCP connections
 *  If set to `v1` the version 1 (human readable) of ProxyProtocol is used
 *  If set to `v2` the version 2 (binary) of ProxyProtocol is used

openapi.yaml
```diff
@@ -17139,6 +17145,25 @@ components:
...
+            text while `v2` makes use of a binary signature.
+
+
+            If ommited from the configuration, the default is `none`.
+          type: string
+          enum:
```
aspsk, line 17161: `none` should come first (compare with other items, e.g., `stickiness`)

PR #318, opened 2020-09-02
[Update] Limited access keys with Object Storage

rsyracuse:


openapi.yaml
```diff
@@ -10692,6 +10692,22 @@ paths:
                   type: string
                   description: The label for this keypair, for display purposes only.
                   example: my-key
+                bucket_access:
+                  type: object
```
victoriaroan, line 10696: This should be `type: array` with `items` being of `type: object` and having the properties you added below. Same for the response value below.

openapi.yaml
```diff
@@ -10692,6 +10692,22 @@ paths:
...
+                  description: Defines this key as a limited access key, and contains values pertaining to permissions. Limited Access Keys block access to all buckets except for those specified within the object.
+                  properties:
+                    cluster:
+                      type: string
+                      description: The datacenter where a bucket with limited access is hosted.
+                      example: ap-south-1
+                    bucket_name:
+                      type: string
+                      description: The unique label of the bucket with limited access.
```
victoriaroan, line 10705: `The unique label of the bucket to which the key should have access.`

openapi.yaml
```diff
@@ -10692,6 +10692,22 @@ paths:
                   type: string
                   description: The label for this keypair, for display purposes only.
                   example: my-key
+                bucket_access:
+                  type: object
+                  description: Defines this key as a limited access key, and contains values pertaining to permissions. Limited Access Keys block access to all buckets except for those specified within the object.
+                  properties:
+                    cluster:
+                      type: string
+                      description: The datacenter where a bucket with limited access is hosted.
```
victoriaroan, line 10701: It would be more accurate to say `The Object Storage cluster where a bucket to which the key should have access is hosted.`

The OBJ cluster label is different than the datacenter/region label (typically just a `-1` appended to the datacenter name.)

openapi.yaml
```diff
@@ -17697,6 +17730,26 @@ components:
           description: This keypair's secret key. **Only returned on key creation**.
           example: '[REDACTED]'
           readOnly: true
+        limited:
+          type: boolean
+          description: Whether or not this key is a limited access key. Will return `false` if this key grants full access to all buckets on the user's account.
+          example: true
+        bucket_access:
```
victoriaroan, line 17737: My other comments about type and property descriptions also apply to this section.

victoriaroan (changes requested):

leslitagordita (approved):

victoriaroan (approved):
Copy changes look good 👍

PR #327, opened 2020-09-15
[Update] Docs UI Links

hzoppetti:
updating links for new docs ui

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -17,7 +17,7 @@ info:
     This document was generated from our OpenAPI Specification.  See the
     <a target="_top" href="https://www.openapis.org">OpenAPI website</a> for more information.
 
-    <a target="_top" href="https://developers.linode.com/api/docs/v4/openapi.yaml">Download the Linode OpenAPI Specification</a>.
+    <a target="_top" href="/docs/api/openapi.yaml">Download the Linode OpenAPI Specification</a>.
 
 
     ## Changelog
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -17,7 +17,7 @@ info:
     This document was generated from our OpenAPI Specification.  See the
     <a target="_top" href="https://www.openapis.org">OpenAPI website</a> for more information.
 
-    <a target="_top" href="https://developers.linode.com/api/docs/v4/openapi.yaml">Download the Linode OpenAPI Specification</a>.
+    <a target="_top" href="/docs/api/openapi.yaml">Download the Linode OpenAPI Specification</a>.
```
leslitagordita, line 20: When I build locally, this is a 404. When I look through the docs site directories, I see this file located here: `_vendor/github.com/linode/linode-api-docs/openapi.yaml`. Where you able to access the file in your local build of new docs? I tried both using the local spec file path and the github repo path.

leslitagordita:
Also, all other link updates look good to me. Just have that open question about the `openapi.yaml` file.

openapi.yaml
leslitagordita: @hzoppetti confirmed she was able to download locally. I rebuilt and am able to download the spec file now.

leslitagordita (approved):

openapi.yaml
nmelehan: So, the `_vendor` path is just where Hugo copies in the API spec file "internally" -- it doesn't map directly onto the URL structure that it ultimately renders to.

I think the issue here is that our implementation of the API in the new UI doesn't actually publish the literal openapi.yaml file. It only publishes the markdown pages that a human would read through (at /docs/api/). In other words, if we were publishing the spec file itself, it would appear in the following output; this command is run from the root of the docs repo in the `feature/new-docs-ui` branch, after having previously run `hugo` (or `hugo75`) to build the site:

```
ls public/api
account                      index.json                   linode-types                 nodebalancers                stackscripts
domains                      key                          longview                     object-storage               support
images                       linode-instances             managed                      profile                      tags
index.html                   linode-kubernetes-engine-lke networking                   regions                      volumes
```

We can ask Bjorn to render the spec file (would just need to make an issue in our issue tracker repo).

PR #332, opened 2020-09-18
Add POST, GET, and DELETE Object Storage SSL Cert

leslitagordita:


openapi.yaml
```diff
@@ -11717,6 +11717,135 @@ paths:
...
+          description: Returns whether this bucket has a corresponding SSL certificate that was uploaded by a user.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ObjectStorageSSLResponse'
+        default:
+          $ref: '#/components/responses/ErrorResponse'
+      x-code-samples:
+      - lang: Shell
+        source: >
```
adammohammed, line 11802: I've just put up a PR to change this a bit. It's kind of a pain to have to strip newlines to fit PEM certs within a json string, so instead we'll be accepting the certificate and key as file inputs, similar to how we accept support ticket attachments.

leslitagordita: @adammohammed I added some info about being able to also upload a .crt file. If this feature doesn't get added, I'll revert the commit.

adammohammed: @leslitagordita Yeah looks like if we add this we'll have to do the same for node balancers. So I think let's revert it for now. Those changes will have to come later.

leslitagordita:
Closing this PR in preference for the hotfix: https://github.com/linode/linode-api-docs/pull/336

PR #349, opened 2020-10-20
Release 4.78.3

hzoppetti:
Updates links for new docs UI


yoshizzle (approved):
Personally, I'm a fan of keeping "v4" in the URL for the future potential that a v5 would happen. There hasn't been any official talk of v5 anytime soon, so removal of v4 isn't a deal-breaker for me. But I do like having it in, fwiw.

PR #351, opened 2020-10-30
Add GET Object ACL

leslitagordita:
* Added GET /object-storage/buckets/{clusterId}/{bucket}/object-acl

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -11280,6 +11280,16 @@ paths:
       - personalAccessToken: []
       - oauth:
         - object_storage:read_only
+      parameters:
+      - name: name
+        in: query
+        required: true
+        description: >
+          The `name` of the object for which to retrieve its Access Control List (ACL).
+          Use the [Object Storage Bucket Contents List](/docs/api/object-storage/#object-storage-bucket-contents-list)
+          endpoint to access all object names in a bucket.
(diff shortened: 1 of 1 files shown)


patthiel:
the endpoint requires a `name` query string parameter. This accepts the name of the object in the bucket you want to get object ACL's for.. For example:

```shell
$ curl object-storage/buckets/alpha/pats_cool_bucket/object-acl?name=test.txt

```

everything else looks good

victoriaroan (changes requested):
`name` is accepted as a query parameter. It is the name of the object and it is required.

openapi.yaml
```diff
@@ -11246,6 +11246,80 @@ paths:
...
+            application/json:
+              schema:
+                type: object
+                properties:
+                  acl:
+                    type: string
+                    enum:
+                    - private
+                    - public-read
+                    - authenticated-read
```
victoriaroan, line 11306: It can also return `public-read-write` and `custom`.

victoriaroan (changes requested):

openapi.yaml
leslitagordita: Updated!

victoriaroan (approved):

PR #353, opened 2020-10-30
Add GET /object-storage/transfer/

leslitagordita:


openapi.yaml
```diff
@@ -11850,6 +11850,46 @@ paths:
...
+        Object Storage adds 1 terabyte of outbound data transfer to your data transfer pool.
+        See the [Object Storage Pricing and Limitations](/docs/guides/pricing-and-limitations/)
+        guide for details on Object Storage transfer quotas.
+
+      tags:
+      - Object Storage
+      security:
+      - personalAccessToken: []
+      - oauth:
+        - object_storage:read_write
```
adammohammed, line 11871: This should be `object_storage:read_only`

openapi.yaml
```diff
@@ -11850,6 +11850,46 @@ paths:
...
+      responses:
+        '200':
+          description: A paginated list of buckets you own.
```
adammohammed, line 11874: Does this need to have a description? The transfer description below should suffice. This endpoint is also not paginated.

adammohammed (changes requested):

openapi.yaml
leslitagordita: Updated!

openapi.yaml
leslitagordita: Our theme requires a description here. I've updated the description copy.

adammohammed (approved):
Looks good, thank you!

PR #377, opened 2021-01-12
Release 4.83.0

bbiggerr:


openapi.yaml
```diff
@@ -6726,6 +6738,50 @@ paths:
...
+      summary: Kubernetes Cluster Recycle
+      description: |
+        Recycles a designated Kubernetes Cluster. All Linodes within the Cluster will be deleted
+        and replaced with new Linodes on a rolling basis, which may take several minutes. Replacement Nodes are
+        installed with the latest available patch for the Cluster's Kubernetes Version.
+
+        **Any local storage on deleted Linodes (such as "hostPath" and "emptyDir" volumes, or "local" PersistentVolumes) will be erased.**
+      responses:
+        '200':
+          description: Cluster has been recycled.
```
sibucan, line 6768: When you get a 200 OK from this endpoint, it doesn't mean the cluster has been recycled, but rather that the request to recycle the cluster has succeeded and is now in progress.

openapi.yaml
```diff
@@ -6517,6 +6517,17 @@ paths:
                     description: >
                       An array of tags applied to the Kubernetes cluster. Tags are for organizational purposes only.
                       To delete a tag, exclude it from your `tags` array.
+                k8s_version:
+                    type: string
+                    description: >
+                      The desired upgraded Kubernetes version for this Kubernetes cluster in the format of
+                      &lt;major&gt;.&lt;minor&gt;. New and recycled Nodes in this cluster will be installed with the
+                      latest available patch for the Cluster's upgraded Kubernetes version.
```
g1bl3t, line 6525: Can we remove the instances of the word "upgraded" in this first paragraph? I think that makes more sense here, since this is representing the desired state of the cluster and that may not necessarily be an upgrade. For example, you can hit this endpoint giving the same version you are currently on.
```suggestion
                      The desired Kubernetes version for this Kubernetes cluster in the format of
                      &lt;major&gt;.&lt;minor&gt;. New and recycled Nodes in this cluster will be installed with the
                      latest available patch for the Cluster's Kubernetes version.
```

openapi.yaml
```diff
@@ -6918,6 +6974,125 @@ paths:
...
+                        example: "12345-6aa78910bc"
+                      instance_id:
+                        type: integer
+                        description: >
+                          The Linode's ID. If no Linode is currently provisioned for this Node, this is `null`.
+                        example: 123456
+                      status:
+                        type: string
+                        description: >
+                          The Node's status as it pertains to being a Kubernetes node.
```
sibucan, line 7031: This status relates more to whether or not the Linode has successfully been created with the necessary Kubernetes software. `not_ready` means that the Linode is still in the process of being created, while `ready` means that the Linode has all the software and is in a Running state. It does _not_ mean that the Kubernetes Node Object (which you can see when you `kubectl get nodes`) itself is ready. We want to add that information and have it be accessible from APIv4 at some point in the future.

openapi.yaml
```diff
@@ -6918,6 +6974,125 @@ paths:
...
+      summary: Node Recycle
+      description: |
+        Recycles an individual Node in the designated Kubernetes Cluster. The Node will be deleted
+        and replaced with a new Linode, which may take a few minutes. Replacement Nodes are
+        installed with the latest available patch for the Cluster's Kubernetes Version.
+
+        **Any local storage on deleted Linodes (such as "hostPath" and "emptyDir" volumes, or "local" PersistentVolumes) will be erased.**
+      responses:
+        '200':
+          description: Node has been recycled.
```
sibucan, line 7079: When you get a 200 OK from this endpoint, it doesn't mean the node has been recycled, but rather that the request to recycle the node has succeeded and is now in progress.

g1bl3t (changes requested):
1 small suggestion on the wording

sibucan (approved):

PR #404, opened 2021-03-08
[New] Service Transfers Endpoints

bbiggerr:
* Add Service Transfers
* Deprecate Entity Transfers

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -348,10 +348,10 @@ info:
 
     * [Open Support Ticket](/docs/api/support/#support-ticket-open)
 
-    Accepting Entity Transfers has a dedicated limit of 2 requests per minute per user.
+    Accepting Service Transfers has a dedicated limit of 2 requests per minute per user.
     That endpoint is:
 
-    * [Entity Transfer Accept](/docs/api/account/#entity-transfer-accept)
+    * [Service Transfer Accept](/docs/api/account/#service-transfer-accept)
 
     ## CLI (Command Line Interface)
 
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -2005,6 +1883,312 @@ paths:
...
+          format: uuid
+    get:
+      x-linode-grant: unrestricted only
+      tags:
+      - Account
+      summary: Service Transfer View
+      description: |
+        Returns the details of the Service Transfer for the provided token.
+
+        This command can only be accessed by the unrestricted users of the accounts that created or accepted this
```
alsaihn, line 2049: This needs a bit more clarification. If the transfer is pending, any unrestricted user may view it. This is because we don't know who the transfer is intended for. In any other state, only the creator/acceptor accounts' unrestricted users may view it.

bbiggerr: Is this clearer?
"Until a transfer is accepted, unrestricted users of the account that created it can view it. After a transfer has been accepted, it can also be viewed by any unrestricted user of the account that accepted the transfer."

yoshizzle: @bbiggerr Carissa means any unrestricted user at all. Meaning any user for any customer.

yoshizzle (approved):
approved once the suggested change from Carissa and me is addressed.

openapi.yaml
bbiggerr: I think this is the clearest way to put it:

While a transfer is pending, any unrestricted user *of any account* can access this command. After a
transfer has been accepted, it can only be viewed by unrestricted users of the accounts that created and
accepted the transfer. If cancelled or expired, only unrestricted users of the account that created the
transfer can view it.

PR #405, opened 2021-03-09
[Update] Firewalls endpoints

bbiggerr:


openapi.yaml
```diff
@@ -9812,25 +9810,22 @@ paths:
...
+              allOf:
+              - $ref: '#/components/schemas/Firewall'
               required:
-                - label
-                - rules
+              - label
+              - rules
+              - inbound_policy
+              - outbound_policy
+              - action
```
alsaihn, line 9820: Do these need to be indented slightly more?

alsaihn (approved):

openapi.yaml
bbiggerr: We don't need to, though it looks like indents in these cases do follow correct yaml conventions. We've been historically inconsistent with this. Would rather fix them all in one go if I'm going to do it.

PR #407, opened 2021-03-10
Fix `linode-cli nodebalancers config-rebuild` command

Dorthu:
This is for linode/linode-cli#226

Looks like the spec for this endpoint was wrong, so the CLI didn't work.
This change fixes the spec to match what the endpoint wants, and in turn
fixes the CLI.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -11240,7 +11240,7 @@ paths:
                 properties:
                   nodes:
                     type: array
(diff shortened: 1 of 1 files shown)


yoshizzle (approved):
May need some formatting for the bullets, but @bbiggerr would know what their convention is now. Content looks good to me.

PR #411, opened 2021-03-19
[Update] Account View GET /account - add service_type to active_promotions

bbiggerr:


alsaihn (approved):

openapi.yaml
```diff
@@ -15388,6 +15388,25 @@ components:
                 description: >
                   The amount of credit left for this month for this promotion.
                 example:  "10.00"
+              service_type:
+                type: string
+                enum:
+                  - all
```
yoshizzle, line 15394: All does not show up in the list here: https://bits.linode.com/LinodeAPI/nimbus/blob/development/nimbus/objects/service.py#L26

is that intentional?

cc @nbokoski @bbiggerr

bbiggerr: Here's the reference that I picked it up from:
https://bits.linode.com/LinodeAPI/nimbus/blob/development/nimbus/objects/promotion.py#L19
https://bits.linode.com/LinodeAPI/nimbus/pull/1187

My understanding is that a `service_type` of `all` in this case is necessary for general promotion credits.

yoshizzle (approved):

yoshizzle:
Cool. I was looking at the wrong enum. :) Lgtm!

PR #419, opened 2021-04-13
Update VLANs schema and endpoint description

bbiggerr:


yoshizzle (approved):

alsaihn (approved):
The detach section is much clearer now, ty!

PR #447, opened 2021-05-17
[Update] GET /images - type filterable

bbiggerr:


openapi.yaml
```diff
@@ -17377,6 +17378,7 @@ components:
           x-linode-cli-display: 6
         type:
           type: string
+          x-linode-filterable: true
           description: >
             How the Image was created. Manual Images can be created at any time.
```
yoshizzle, line 17383: To be consistent with the way manual is described above.

```suggestion
            How the Image was created. "Manual" Images can be created at any time.
```

yoshizzle (changes requested):

alsaihn (approved):

PR #468, opened 2021-06-21
Add contentEncoding and contentMediaType for Kubeconfig

Dorthu:
This is related to https://github.com/linode/linode-cli/issues/247

The kubeconfig returned from `GET /lke/clusters/{clusterId}/kubeconfig`
is a base64-encoded YAML config.  The OpenAPI standard now supports this
[with the standard JSON Schema fields](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#considerations-for-file-uploads).

This change notes the encoding of the `kubeconfig` field, which the CLI
may use to decode the base64 data and display the actual YAML for
convenience (see linked CLI PR below).

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -8272,6 +8272,8 @@ paths:
                     readOnly: true
                     description: >
                       The Base64-encoded Kubeconfig file for this Cluster.
+                    contentEncoding: base64
+                    contentMediaType: application/yaml
         default:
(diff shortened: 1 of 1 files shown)


Dorthu:
This will not pass validation without https://github.com/Dorthu/openapi3/pull/48, as these fields were previously not included in the parser/validator.

PR #472, opened 2021-06-24
[New] Payment Method Add (POST /account/payment-methods)

bbiggerr:
* Deprecate Credit Card Add/Edit (POST /account/credit-card)

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -16048,7 +16048,7 @@ components:
           maxLength: 4
           format: digits
           description: >
-            CVV (Card Verification Value) of the credit card, typically found on the back of the card. **** 
+            CVV (Card Verification Value) of the credit card, typically found on the back of the card.
           example: '123'
     Device:
       type: object
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -543,6 +543,45 @@ paths:
         source: >
           linode-cli account cancel \
             --comments "I'm consolidating my accounts"
+  /account/client-token:
+    x-linode-cli-command: account
+    get:
+      servers:
+        - url: https://api.linode.com/v4
+      x-linode-grant: read_only
+      tags:
+        - Account
+      summary: Payment Client Token View
+      description: |
+        Generates a Payment Client Token.
+      operationId: getClientToken
+      x-linode-cli-action: client-token
+      security:
+        - personalAccessToken: []
+        - oauth:
+          - account:read_only
+      responses:
+        '200':
+          description: Client Token generated.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  client_token:
+                    type: string
+                    description: The Payment Client Token.
+                    example: 1234
+        default:
+          $ref: '#/components/responses/ErrorResponse'
+      x-code-samples:
+        - lang: Shell
+          source: >
+            curl -H "Authorization: Bearer $TOKEN" \
+                https://api.linode.com/v4/account/client-token
+        - lang: CLI
+          source: >
+            linode-cli account client-token
   /account/credit-card:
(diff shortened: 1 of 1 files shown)


nbokoski (approved):

openapi.yaml
```diff
@@ -551,15 +551,9 @@ paths:
...
+        **DEPRECATED**. Please use Payment Method Add ([POST /account/payment-methods](/docs/api/account/#payment-method-add)).
 
-        Only one credit card can be associated with your Account, so using this
-        endpoint will overwrite your currently active card information with the
-        new credit card.
-
-        To use this endpoint, you must have a valid `zip` entered for your Account.
-        Use the Account Update ([PUT /account](/docs/api/account/#account-update))
-        endpoint to enter a new zip code.
+        Adds a credit card Payment Method to your account and sets it as the default method.
```
yoshizzle, line 556: In this comment, it is stated "and sets it as the default method"

However, in another edit below, it states "with the option to set it as the default method."

This should be clarified to the correct statement in whichever place is incorrect.

openapi.yaml
```diff
@@ -1732,6 +1679,82 @@ paths:
...
             linode-cli account payment-methods-list
+    post:
+      servers:
+        - url: https://api.linode.com/v4
+      x-linode-grant: read_write
+      tags:
+        - Account
+      summary: Payment Method Add
+      description: |
+        Adds a Payment Method to your Account with the option to set it as the default method.
```
yoshizzle, line 1690: Here is the other line from the comment above

openapi.yaml
```diff
@@ -16006,14 +16032,24 @@ components:
...
             must result in a month/year combination of the current month or in
             the future. An expiration date set in the past is invalid.
           example: 2020
+        cvv:
+          type: string
+          minLength: 3
+          maxLength: 4
+          format: digits
+          description: >
+            CVV (Card Verification Value) of the credit card, typically found on the back of the card. ****
```
yoshizzle, line 16051: What are the four asterisks here for?

yoshizzle (changes requested):

openapi.yaml
bbiggerr: This describes the behavior of the deprecated `PUT /account/credit-card` endpoint, which takes in a different request body that doesn't include an `is_default` field, so it currently sets the payment method as the default automatically in the backend.

openapi.yaml
bbiggerr: A placeholder that I forgot to remove! Removing now...

openapi.yaml
```diff
@@ -551,15 +551,9 @@ paths:
...
-        Adds/edit credit card information to your Account.
+        **DEPRECATED**. Please use Payment Method Add ([POST /account/payment-methods](/docs/api/account/#payment-method-add)).
 
-        Only one credit card can be associated with your Account, so using this
-        endpoint will overwrite your currently active card information with the
-        new credit card.
-
-        To use this endpoint, you must have a valid `zip` entered for your Account.
-        Use the Account Update ([PUT /account](/docs/api/account/#account-update))
-        endpoint to enter a new zip code.
```
alsaihn, line 562: This bit is still valid and this endpoint will still report an error if the listed zip is not valid. Should we keep it?

yoshizzle (approved):

alsaihn (approved):

openapi.yaml
bbiggerr: I reworded this a bit in the Payment Method Add description to be a bit broader. Agreed that this bit is still valid -- updating that description to be specific about requiring a valid `zip` as well.

PR #485, opened 2021-07-12
[Update] DomainRecord schema for null MX records

bbiggerr:


openapi.yaml
```diff
@@ -16431,31 +16431,36 @@ components:
...
-            The name of this Record. For requests, this property's actual usage and whether it is required depends on the type of record this represents:
+            The name of this Record. For requests, this property's actual usage and whether it is required depends
+            on the type of record this represents:
 
             `A` and `AAAA`: The hostname or FQDN of the Record.
 
             `NS`: The subdomain, if any, to use with the Domain of the Record.
 
-            `MX`: The subdomain.
+            `MX`: The mail server subdomain. To create a [Null MX record](https://datatracker.ietf.org/doc/html/rfc7505),
```
yoshizzle, line 16441: May I suggest changing the order of these instructions? I would maybe say something like:

```suggestion
            `MX`: The mail server subdomain. To create a [Null MX record](https://datatracker.ietf.org/doc/html/rfc7505), first remove any additional MX records, then enter an empty string for this property.
```

With the appropriate links and parenthetical info, etc.

yoshizzle (changes requested):

openapi.yaml
```diff
@@ -16480,14 +16490,18 @@ components:
           type: integer
           minimum: 0
           maximum: 255
-          description: >
-            The priority of the target host for this Record. Lower values are preferred. Only valid and required for SRV record requests.
+          description: |
+            The priority of the target host for this Record. Lower values are preferred. Only valid for
+            MX and SRV record requests. Required for SRV record requests.
+
+            Defaults to `0` if omitted/null for MX record requests.
```
alsaihn, line 16497:
```suggestion
            Defaults to `0` if omitted for Null MX record requests.
```

chris-sarli, line 16497: I first read "null for MX record requests" as related to null MX records, but it makes sense after rereading.

For clarity, maybe add a sentence saying that the `priority` must be `0` for null MX records?

alsaihn (approved):

openapi.yaml
```diff
@@ -16480,14 +16490,18 @@ components:
           type: integer
           minimum: 0
           maximum: 255
-          description: >
-            The priority of the target host for this Record. Lower values are preferred. Only valid and required for SRV record requests.
+          description: |
+            The priority of the target host for this Record. Lower values are preferred. Only valid for
+            MX and SRV record requests. Required for SRV record requests.
+
+            Defaults to `0` for MX record requests. Must be `0` for Null MX records.
```
chris-sarli, line 16497: This is correct, and led me to discover a bug. Currently, if the priority isn't provided, then we reject the request before defaulting to `0`. Putting together a fix now; the documentation describes the correct behavior.

chris-sarli (approved):

PR #527, opened 2021-09-20
Add new properties to Kernels and Images

bbiggerr:


openapi.yaml
```diff
@@ -3750,14 +3750,17 @@ paths:
         Returns a paginated list of Images.
 
         * **Public** Images have IDs that begin with "linode/". These distribution images are generally available to
-        all users. To view only public Images, call this endpoint without authentication.
+        all users. To view only public Images, call this endpoint with or without authentication.
 
         * **Private** Images have IDs that begin with "private/". These images are account-specific and only
-        accessible to users with `images:read_only` authorization. To view private Images you have access to in
-        addition to public images, call this endpoint with authentication.
+        accessible to users with `images:read_only` authorization. To view private Images, call this endpoint with authentication.
```
alsaihn, line 3756: The two "To view..." statements here seem confusing, especially if read in isolation. Maybe move them to their own bullet and flip the content? 

Call this endpoint with authentication to view both public and private images you have access to. You may call this endpoint without authentication to view only public images.

alsaihn (approved):
Not sure if it's a good suggestion or not XD

openapi.yaml
bbiggerr: Love the suggestion -- made edits for clarification below

PR #542, opened 2021-10-05
Update Image Upload CLI settings

bbiggerr:


openapi.yaml
```diff
@@ -3909,18 +3911,17 @@ paths:
...
       - lang: CLI
         source: >
-          linode-cli image-upload \
+          linode-cli images upload \
+            --description "Optional details about the Image" \
             --label "Example Image" \
-            --description "Some details about the image" \
-            --region us-east \
-            ~/Downloads/image-file.img.gz
+            --region us-east
```
Dorthu, line 3924: Personally, I'd still leave the plugin invocation in the example here - it's more of the correct way to do it.

Dorthu (approved):

openapi.yaml
bbiggerr: Thanks for the suggestion -- went with including both commands

PR #562, opened 2021-11-15
[New] IPv6 Range commands

bbiggerr:


yoshizzle (approved):
lgtm

just one thought. The /networking/ipv4/assign technically still exists - it won't be officially deprecated for a little bit. Not sure if that affects your replacing of the existing "ipv4" endpoint with the new "ips" one.

bbiggerr:
@yoshizzle the "replacement" is from the customer's perspective when viewing the docs -- I've kept the old command in the same path, just renamed it with a reference to the new command so we should be good

openapi.yaml
```diff
@@ -10185,20 +10185,126 @@ paths:
...
       summary: Linodes Assign IPs
+      description: |
+        Assign multiple IPv4 addresses and/or IPv6 ranges to multiple Linodes in one Region. This allows
+        swapping, shuffling, or otherwise reorganizing IPs to your Linodes.
+
+        The following restrictions apply:
+        * All Linodes involved must have at least one public IPv4 address after assignment.
+        * Linodes may have no more than one assigned private IPv4 address.
+        * Linodes may have no more than one assigned IPv6 range.
+        * [Open a Support Ticket](/docs/api/support/#support-ticket-open) to request expansion of these restrictions.
```
alsaihn, line 10203: I don't believe the first restriction can be altered, would it be better to be more specific here?
```suggestion
        * [Open a Support Ticket](/docs/api/support/#support-ticket-open) to request additional IPv4 addresses or IPv6 ranges.
```

jdawley (approved):

openapi.yaml
bbiggerr: yes, that's definitely clearer -- nice catch

alsaihn (approved):

PR #569, opened 2021-11-29
[Update] Object Storage Cancel; Object Storage Bucket Remove

bbiggerr:


openapi.yaml
```diff
@@ -12561,14 +12560,15 @@ paths:
...
       - url: https://api.linode.com/v4
       summary: Object Storage Bucket Remove
-      description: >
-        Removes a single bucket. While buckets containing objects _may_ be
-        deleted by including the `force` option in the request, such operations
-        will fail if the bucket contains too many objects. The recommended
+      description: |
+        Removes a single bucket.
+
+        Bucket objects must be removed prior to removing the bucket. While buckets containing objects _may_ be
```
yoshizzle, line 12566: Do bucket objects need to be removed first? I thought I saw some chatter somewhere in Slack that indicated you did not need to remove first. You should confirm.

openapi.yaml
```diff
@@ -12561,14 +12560,15 @@ paths:
...
+        deleted using the [s3cmd command-line tool](/docs/products/storage/object-storage/guides/s3cmd/#delete-a-bucket), such operations
+        can fail if the bucket contains too many objects. The recommended
```
bcmertz, line 12568: "Bucket objects must be removed prior to removing the bucket" - This will no longer be true once bapi support is merged and the API force flag is deprecated.

openapi.yaml
bcmertz: You're correct @yoshizzle

bcmertz:
This is blocked until https://bits.linode.com/LinodeAPI/apinext/pull/4002 and it's needed bapi changes are merged

openapi.yaml
bcmertz: Update: after speaking to @bbiggerr the changes here only reflect  https://bits.linode.com/LinodeAPI/apinext/pull/4001 and remove reference to the force flag (prior to it being removed). With this in mind ignore the above comment :)

bcmertz (approved):

PR #205, opened 2020-02-20
[DNM] Bare Metal

hzoppetti:
start of bare metal work

PR #574, opened 2021-12-01
[DNM] changes for Bare Metal Beta support

jamesottinger:
Bare Metal Beta support change set. Replacing the older #205 version.


Dorthu (approved):
lgtm :+1:

PR #576, opened 2021-12-14
[New] POST /networking/ips/share; GET /networking/ipv6/ranges/{range}

bbiggerr:


openapi.yaml
```diff
@@ -10307,6 +10307,94 @@ paths:
             --region us-east \
             --assignments.address 2001:db8:3c4d:15::/64 \
             --assignments.linode_id 123
+  /networking/ips/share:
+    x-linode-cli-command: networking
+    post:
+      servers:
+      - url: https://api.linode.com/v4beta
```
yoshizzle, line 10314: this isn't in beta

openapi.yaml
```diff
@@ -10307,6 +10307,94 @@ paths:
...
+        the primary Linode becomes unresponsive. This means that requests to the primary Linode's IP address can be
+        automatically rerouted to secondary Linodes at the configured shared IP addresses.
+
+        IP failover requires configuration of a failover service (such as [Keepalived](/docs/guides/ip-failover-keepalived))
+        within the internal system of the primary Linode.
+
+        **Note**: IPv6 range sharing has limited availability in certain regions. Please contact customer support for
+        assistance in enabling IPv6 range sharing for your Linodes.
+
+        **Beta**: This endpoint is in beta. Please make sure to prepend all requests with
```
yoshizzle, line 10332: Same as above. This isn't in beta so far as I know.

yoshizzle (changes requested):

openapi.yaml
```diff
@@ -10307,6 +10307,94 @@ paths:
...
+        `/v4beta` instead of `/v4`, and be aware that this endpoint may receive breaking
+        updates in the future. This notice will be removed when this endpoint is out of
+        beta.
+      operationId: shareIPs
+      x-linode-cli-action: ip-share
```
alsaihn, line 10337: This is the same cli action as the older v4 share endpoint. Is that going to cause issues?

openapi.yaml
bbiggerr: No problem reverting, but I'm getting {"errors": [{"reason": "Method Not Allowed"}]} when trying to hit the v4 server

bbiggerr: added v4 server in the next line and removed beta status references -- beta tag still shows up in the docs at the moment, but the CLI won't work with the command until the v4 server issue is fixed for this command otherwise

yoshizzle (approved):

PR #578, opened 2021-12-14
Update NodeBalancer hostname

bbiggerr:


yoshizzle (approved):
would check with @jfrederickson or @stvnjacobs to ensure this is a valid example.

openapi.yaml
```diff
@@ -20248,8 +20248,8 @@ components:
         hostname:
           type: string
           description: >
-            This NodeBalancer's hostname, ending with _.nodebalancer.linode.com_
-          example: nb-207-192-68-16.newark.nodebalancer.linode.com
+            This NodeBalancer's hostname, ending with _.ip.linodeusercontent.com_
```
alsaihn, line 20251:
```suggestion
            This NodeBalancer's IPv4 address, ending with _.ip.linodeusercontent.com_
```

alsaihn (approved):

openapi.yaml
```diff
@@ -20248,8 +20248,8 @@ components:
...
+          example: 123-456-789-00.ip.linodeusercontent.com
```
jfrederickson, line 20252: Nitpick: that's not a valid IPv4 address, so this hostname could never be generated.

I would probably go with one of the IPs from a range reserved for documentation: https://datatracker.ietf.org/doc/html/rfc5737

PR #581, opened 2021-12-24
bug: Add missing CLI commands/actions to operations

Dorthu:
This is for the Linode CLI.

In the current CLI release, we see this:

```bash
wsmith@linode::linode-cli[master]$ linode-cli default --help
linode-cli default [ACTION]

Available actions:
┌──────────────────────┬────────────────────────┐
│ action               │ summary                │
├──────────────────────┼────────────────────────┤
│ getEntityTransfers   │ Entity Transfers List  │
│ createEntityTransfer │ Entity Transfer Create │
│ getEntityTransfer    │ Entity Transfer View   │
│ deleteEntityTransfer │ Entity Transfer Cancel │
│ acceptEntityTransfer │ Entity Transfer Accept │
│ import               │ Domain Import          │
└──────────────────────┴────────────────────────┘
```

These operations appear in "default" because they have no
`x-linode-cli-command` defined in the spec (so they fall into the
default command).  This isn't ideal.  Additionally, many of those
commands are using their operationId as their action since they lack a
`x-linode-cli-action` element.

This change adds a command and action to all of the above.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -598,9 +598,11 @@ paths:
               --expiry_year 2025 \
               --cvv 111
   /account/entity-transfers:
+    x-linode-cli-command: account
     get:
       deprecated: true
       x-linode-grant: unrestricted only
+      x-linode-cli-action: entity-transfers-list
       parameters:
       - $ref: '#/components/parameters/pageOffset'
       - $ref: '#/components/parameters/pageSize'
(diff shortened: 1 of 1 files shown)


alsaihn (approved):

bbiggerr (commented):
Entity Transfer endpoints are deprecated -- should commands have `x-linode-cli-skip: true`?

Dorthu:
I didn't realize these were renamed.  I'll mark them as skipped instead.

Dorthu:
Opened an associated CLI PR to remove the "default" command from the CLI, which otherwise would be empty (as all operations in it were skipped)

bbiggerr (approved):

PR #584, opened 2022-01-14
bug: Fix OAuth Exchange docs

Dorthu:
This was reported by a Customer.

When performing a private OAuth Exchange, the correct value for
`grant_type` is "authorization_code".  This parameter may also be
omitted, as this value is the default; it only need be provided (as
"refresh_token") when doing an OAuth Token Refresh.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -126,14 +126,14 @@ info:
     Authorization: Bearer 03d084436a6c91fbafd5c4b20c82e5056a2e9ce1635920c30dc8d81dc7a6665c
     ```
 
-    To refresh access, make another request with the following parameters int he POST body:
+    This `access_token` is set to expire in two hours. To refresh access prior to expiration, make another request to the same URL with the following parameters in the POST body:
 
     | PARAMETER | DESCRIPTION |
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -113,19 +112,31 @@ info:
...
 
-    Included in the reponse is an `access_token`. With this token, you can proceed to make
+    Included in the response is an `access_token`. With this token, you can proceed to make
     authenticated HTTP requests to the API by adding this header to each request:
 
     ```
     Authorization: Bearer 03d084436a6c91fbafd5c4b20c82e5056a2e9ce1635920c30dc8d81dc7a6665c
     ```
 
+    To refresh access, make another request with the following parameters int he POST body:
```
Dorthu, line 129:
```suggestion
    To refresh access, make another request to the same URL with the following parameters int he POST body:
```

It might also be worth explaining why you'd want to refresh a token (which is to continue an existing session once the originally-generated token expires, usually after 2 hours).

bbiggerr (approved):

yoshizzle (approved):

PR #592, opened 2022-02-03
[New] Managed Databases

bbiggerr:


yoshizzle (changes requested):

openapi.yaml
```diff
@@ -3062,6 +3062,855 @@ paths:
...
+      summary: Managed MySQL Database Backup Restore
+      operationId: postDatabasesMySQLInstanceBackupRestore
+      servers:
+      - url: https://api.linode.com/v4beta
+      x-linode-cli-action: mysql-backup-restore
+      x-linode-grant: unrestricted only
+      description: |
+        Restore a backup to an active Managed MySQL Database on your Account.
+
+        Only unrestricted Users can access this command.
```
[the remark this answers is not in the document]
bbiggerr: Definitely a benefit if it makes it clearer beyond stating "an active Managed MySQL Database". I'll move the requirement to its own line so that it stands out more on all commands that require an active state.

openapi.yaml
```diff
@@ -3062,6 +3062,881 @@ paths:
...
+      summary: Managed Database Type View
+      operationId: getDatabasesType
+      servers:
+      - url: https://api.linode.com/v4beta
+      x-linode-cli-action: type-view
+      parameters:
+      - $ref: '#/components/parameters/pageOffset'
+      - $ref: '#/components/parameters/pageSize'
+      description: |
+        Display the details of a single Managed Database type. The type determines the resources and price of a Managed Database instance.
```
alsaihn, line 3903:
```suggestion
        Display the details of a single Managed Database type. The type and number of nodes determine the resources and price of a Managed Database instance.
```
Is this better here?

openapi.yaml
```diff
@@ -3062,6 +3062,857 @@ paths:
...
+      summary: Managed Database Engine View
+      operationId: getDatabasesEngine
+      servers:
+      - url: https://api.linode.com/v4beta
+      x-linode-cli-action: engine-view
+      parameters:
+      - $ref: '#/components/parameters/pageOffset'
+      - $ref: '#/components/parameters/pageSize'
+      description: |
+        Display information for a single Managed Database engine type.
```
yoshizzle, line 3130:
```suggestion
        Display information for a single Managed Database engine type and version.
```

openapi.yaml
```diff
@@ -3062,6 +3062,855 @@ paths:
...
+      servers:
+      - url: https://api.linode.com/v4beta
+      x-linode-cli-action: mysql-update
+      x-linode-grant: unrestricted only
+      description: |
+        Update an active Managed MySQL Database.
+
+        Only unrestricted Users can access this command.
+
+        Updating addresses in the `allow_list` overwrites any existing addresses.
```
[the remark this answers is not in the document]
yoshizzle: empty array `[]` blocks everything

specific IPs obviously only allows those IPs

I don't think we want to publicly tell people about 0.0.0.0/0 so perhaps leave that out for now

openapi.yaml
```diff
@@ -3062,6 +3062,857 @@ paths:
...
+      - url: https://api.linode.com/v4beta
+      x-linode-cli-action: mysql-backup-restore
+      x-linode-grant: unrestricted only
+      description: |
+        Restore a backup to a Managed MySQL Database on your Account.
+
+        The Database must have an `active` status to perform this command.
+
+        Only unrestricted Users can access this command.
+
```
yoshizzle, line 3624: Here's a tricky one, Ben. Currently - SG is aware of this - there is a bug with restoring from backup if the customer has reset their credentials after the last backup. It will result in a failed cluster that we need to _try_ and fix.

Maybe you can put something in here about that? Hopefully it's temporary, and maybe this should be run up the stakeholder chain to see if we want to make it known publicly or not.

openapi.yaml
```diff
@@ -3062,6 +3062,857 @@ paths:
...
+      tags:
+      - Databases
+      summary: Managed MySQL Database Credentials Reset
+      operationId: postDatabasesMySQLInstanceCredentialsReset
+      servers:
+      - url: https://api.linode.com/v4beta
+      x-linode-cli-action: mysql-creds-reset
+      x-linode-grant: unrestricted only
+      description: |
+        Reset the root password for a Managed MySQL Database.
```
yoshizzle, line 3719: We may want to add a note here that it takes a few seconds for the credentials to reset and show the new creds.

openapi.yaml
```diff
@@ -16680,6 +17530,383 @@ components:
...
+          x-linode-cli-display: 5
+        hosts:
+          type: object
+          readOnly: true
+          description: The primary and secondary hosts for the Managed Database. These are assigned after provisioning is complete.
+          properties:
+            primary:
+              type: string
+              description: The primary host for the Managed Database.
+              example: lin-123-456-mysql-master.servers.staging.linodedb.net
```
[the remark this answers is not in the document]
yoshizzle: We did change master, so you should change this.

yoshizzle (changes requested):
There was also a comment in the go/no-go about being explicit to customers that the secondary hostname (for private IP/within DC traffic) is publically viewable/accessible? I think Matt has more info on that but it should absolutely be included.

yoshizzle (approved):

PR #701, opened 2022-10-26
Update NodeBalancer and OBJ SSL cert guidance and examples

bbiggerr:


openapi.yaml
```diff
@@ -14404,22 +14407,52 @@ paths:
...
+                      ssl_cert:
+                        $ref: '#/components/schemas/NodeBalancerConfig/properties/ssl_cert'
+                      ssl_key:
+                        $ref: '#/components/schemas/NodeBalancerConfig/properties/ssl_key'
+                      nodes:
+                        type: array
+                        description: |
+                          The NodeBalancer Node(s) that serve this Config.
+                        items:
+                          $ref: '#/components/schemas/NodeBalancerNode'
```
bbiggerr, line 14455: This is expanded because `items` defined with `allOf` currently don't render on the docs site. Can be reverted after a fix is made, which I've already requested.

alsaihn (approved):
mmmm hyphens.

PR #720, opened 2022-11-28
Update Profile SSH Key Add/Update with key types

bbiggerr:
* Specifies valid SSH key types
* Fixes command examples


yoshizzle (approved):
Same for @LBGarber to review

PR #721, opened 2022-11-28
Fix linode-cli phone command actions

bbiggerr:


yoshizzle (approved):
@LBGarber should probably look at this as well

lgarber-akamai (approved):

PR #732, opened 2023-01-09
Account Maintenance List out of beta

bbiggerr:


afolksetapart (approved):

openapi.yaml
```diff
@@ -24003,10 +23996,16 @@ components:
           x-linode-filterable: true
           type: string
           enum:
-          - pending
```
alsaihn, line 24006: We combine ticketing, ticketed, scheduled, and ready statuses into 'pending', so the end user will only see pending, started, and completed.

PR #821, opened 2023-07-07
Mark `nodes` in nodebalancer config as not required

zliang-akamai:
I tried an API call and it doesn't seem to be required.

```
curl -H "Content-Type: application/json" \
    -H "Authorization: Bearer $TOKEN" \
    -X POST -d '{"region": "us-central"}' https://api.linode.com/v4/nodebalancers
```


zliang-akamai:
My bad, didn't realize that was a required for a config, not for the whole API call. Closing this PR

PR #822, opened 2023-07-14
IP Addresses List - added filterable properties

bbiggerr:
* Creates new schema for IP Addresses List response to isolate filterable properties to the command
* Removes filterable tags from IP Address schema that are used in commands that don't provide for filtering

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1,6 +1,6 @@
 openapi: 3.0.1
 info:
-  version: 4.157.0
+  version: 4.157.1
 
   title: Linode API
   description: |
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -21550,6 +21538,99 @@ components:
...
+                    example: test.example.org
+                    nullable: true
+                  linode_id:
+                    type: integer
+                    description: >
+                      The ID of the Linode this address currently belongs to. For IPv4
+                      addresses, this is by default the Linode that this address was
+                      assigned to on creation, and these addresses my be moved using the
+                      [/networking/ipv4/assign](/docs/api/networking/#ips-to-linodes-assign)
+                      endpoint. For SLAAC and link-local addresses, this value may not be changed.
```
alsaihn, line 21622:
```suggestion
                      The ID of the Linode this address currently belongs to. For IPv4
                      addresses, this defaults to the Linode that this address was
                      assigned to on creation. IPv4 addresses may be moved using the
                      [/networking/ipv4/assign](/docs/api/networking/#ips-to-linodes-assign)
                      endpoint. For SLAAC and link-local addresses, this value may not be changed.
```
Bit of a run-on sentence there, and a typo. Is that better?

alsaihn (approved):

PR #898, opened 2023-12-19
Fix `x-linode-cli-action` Style and Server URLs for Account Availability Endpoints 

zliang-akamai:
- Update CLI actions from Camel case (e.g. `getAvailability`) to Hyphen-cased (e.g. `get-availability `)
- Fix server URL so the CLI will be able to target the endpoints in `v4beta` path.


yec-akamai:
Verified on CLI that this change works: 
image

jddocs (approved):

PR #904, opened 2024-01-12
Update new migration types

andystevensname:
Added new properties to resize and migrate endpoints to reflect new cold / warm values.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -8960,7 +8960,7 @@ paths:
                     will not be performed.
                   example: false
                   default: false
-                migration_type:
+                type:
                   type: string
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -8960,6 +8960,15 @@ paths:
                     will not be performed.
                   example: false
                   default: false
+                migration_type:
```
alsaihn, line 8963:
```suggestion
                type:
```
This is just "type" at this endpoint, the other endpoint is correct with "migration_type". Can we add a note here or above that warm is not an available type for cross DC migrations? It's currently only available for pending migrations.

PR #905, opened 2024-01-12
Release 4.171.0

andystevensname:
Added a new property to two endpoints that determine migration type:

/linode/instances/:id/resize
- Added `migration_type`

/linode/instances/{linodeId}/migrate
- Added `type`

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -8960,6 +8960,20 @@ paths:
                     will not be performed.
                   example: false
                   default: false
+                type:
+                  type: string
+                  enum:
+                  - warm
+                  - cold
+                  description: |
+                    Type of migration used in moving to a new host or Linode type.
+
+                    `warm`: the Linode will not power down until the migration is complete.
+
+                    `cold`: the Linode will be powered down and migrated. When the migration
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -9451,6 +9466,21 @@ paths:
...
+                migration_type:
+                  type: string
+                  enum:
+                  - warm
+                  - cold
+                  description: |
+                    Type of migration used in moving to a new host or Linode type.
+
+                    `warm`: the Linode will not power down until the migration is complete.
+                    Warm migrations are not available for DC migrations.
```
alsaihn, line 9478:
```suggestion

```
That note isn't needed here, a resize never crosses DCs.

openapi.yaml
```diff
@@ -1,6 +1,6 @@
 openapi: 3.0.1
 info:
-  version: 4.170.0
+  version: 4.171.0
```
alsaihn, line 3:
```suggestion
  version: 4.170.2
```
This has to match the API version right?

andystevensname: We're going to open a new PR with the right spec number

PR #913, opened 2024-01-30
Updates to the API for VPC GA

Vernholio:
Made several changes to the openapi.yaml file for the GA release of VPC.


openapi.yaml
```diff
@@ -13333,7 +13328,7 @@ paths:
                   "2001:db8:3c4d:15::"
                 ]
               }' \
-              https://api.linode.com/v4beta/networking/ips/share
```
alsaihn, line 13336: This is not a VPC endpoint and is not leaving beta at this time.

alsaihn (changes requested):
There's one unneeded beta removal here, and a comment in slack for you.

PR #916, opened 2024-02-07
Added scale-up functionality for MySQL and PostgreSQL DBaaS clusters

Vernholio:
Updates include:

* Added new _type_ object to #/components/schemas/DatabaseMySQL/properties/updates. Covers re-sizing of your cluster to a Linode Type with more disk space.
* Added new _type_ object to #/components/schemas/DatabasePostgreSQL/properties/updates. Covers re-sizing of your cluster to a Linode Type with more disk space.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3862,11 +3862,11 @@ paths:
                 type:
                   type: string
                   description: |
-                    Request re-sizing of your cluster to a higher plan by specifying a larger [Linode Type](/docs/api/linode-types/#types-list).
+                    Request re-sizing of your cluster to a higher plan by specifying a larger Linode Type.
 
                     * Needs to be a higher plan than your current Linode Type.
 
-                    * You can't resize from a shared CPU Linode Type to a dedicated CPU, and vice-versa.
+                    * Resizing to a larger Linode Type can accrue additional cost. Review the `price` output in the [Types List](/docs/api/linode-types/#types-list) operation for more information.
 
                     * You can't update the `allow-list` and set a new `type` in the same request.
 
@@ -4661,11 +4661,11 @@ paths:
                 type:
                   type: string
                   description: |
-                    Request re-sizing of your cluster to a higher plan by specifying a larger [Linode Type](/docs/api/linode-types/#types-list).
+                    Request re-sizing of your cluster to a higher plan by specifying a larger Linode Type.
 
                     * Needs to be a higher plan than your current Linode Type.
 
-                    * You can't resize from a shared CPU Linode Type to a dedicated CPU, and vice-versa.
+                    * Resizing to a larger Linode Type can accrue additional cost. Review the `price` output in the [Types List](/docs/api/linode-types/#types-list) operation for more information.
 
                     * You can't update the `allow-list` and set a new `type` in the same request.
 
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3862,13 +3862,13 @@ paths:
                 type:
                   type: string
                   description: |
-                    Request re-sizing of your cluster to a higher plan by specifying a larger Linode Type.
+                    Request re-sizing of your cluster to a Linode Type with more disk space. For example, you could request a Linode Type that uses a higher plan.
 
-                    * Needs to be a higher plan than your current Linode Type.
+                    * Needs to be a Linode Type with more disk space than your current Linode.
 
                     * Resizing to a larger Linode Type can accrue additional cost. Review the `price` output in the [Types List](/docs/api/linode-types/#types-list) operation for more information.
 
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -3859,6 +3859,19 @@ paths:
...
                 updates:
                   $ref: '#/components/schemas/DatabaseMySQL/properties/updates'
+                type:
+                  type: string
+                  description: |
+                    Request re-sizing of your cluster to a higher plan by specifying a larger [Linode Type](/docs/api/linode-types/#types-list).
+
+                    * Needs to be a higher plan than your current Linode Type.
+
+                    * You can't resize from a shared CPU Linode Type to a dedicated CPU, and vice-versa.
```
akasklock, line 3869: This is not accurate -- switching between shared/dedicated _is_ supported.  (One slightly esoteric caveat: switching from a dedicated plan to a shared plan with the same specs is not currently supported, but that's also not intentional.  Switching in the opposite direction is supported, though.)

Vernholio: Thanks, @akasklock ... Just to confirm: By "same specs," are you saying that you can't switch from a 1GB dedicated Linode to a 1GB shared Linode? This wouldn't be "upscaling" then, and it's covered by the point, "Needs to be a higher plan..."

So, I should just drop this whole point, right?

akasklock: I think that's mostly correct, albeit with a couple caveats.

* Switching from, e.g., 1GB shared to 1GB dedicated is supported.  If the median reader would understand that to be "upscaling", then I think dropping this bullet is sufficient.
* We're hoping to remove the limitation on like-to-like switching from dedicated to shared very soon (I'd characterize it as a bug).  I don't know if that has any implications for this PR, though -- perhaps it's just guidance that we're likely to request another small tweak to this verbiage in the near future.

openapi.yaml
```diff
@@ -3859,6 +3859,19 @@ paths:
                   $ref: '#/components/schemas/DatabaseMySQLRequest/properties/allow_list'
                 updates:
                   $ref: '#/components/schemas/DatabaseMySQL/properties/updates'
+                type:
```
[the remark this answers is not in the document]
Vernholio: Good point. I've added a new bullet to address the increased cost and linked to the "Types List operation where customers can review Linode Types and their cost.

openapi.yaml
Vernholio: Thanks a lot for the explanation. I've removed the point for now. We can tweak these docs later when the fix is in place.

alsaihn (approved):

openapi.yaml
```diff
@@ -3859,6 +3859,19 @@ paths:
...
+                  type: string
+                  description: |
+                    Request re-sizing of your cluster to a higher plan by specifying a larger Linode Type.
```
bbiggerr, line 3865: The constraint is that the new Type can't have less disk space than the current Type. The term "higher plan" isn't entirely accurate here.

openapi.yaml
```diff
@@ -3859,6 +3859,19 @@ paths:
...
+
+                    * Needs to be a higher plan than your current Linode Type.
+
+                    * Resizing to a larger Linode Type can accrue additional cost. Review the `price` output in the [Types List](/docs/api/linode-types/#types-list) operation for more information.
+
+                    * You can't update the `allow-list` and set a new `type` in the same request.
```
bbiggerr, line 3871: Should be `allow_list`

openapi.yaml
```diff
@@ -3879,6 +3892,7 @@ paths:
                     "203.0.113.1",
                     "192.0.1.0/24"
                   ],
+                  "type": "g6-standard-1",
                   "updates" = {
```
bbiggerr, line 3896: Noticed an old typo here, should be `"updates": {`

openapi.yaml
```diff
@@ -4663,6 +4691,7 @@ paths:
...
```
bbiggerr, line 4695: Same type -- should be `"updates": {`

openapi.yaml
Vernholio: Updated the statement to accommodate.

openapi.yaml
Vernholio: Fixed.

openapi.yaml
Vernholio: Fixed.

openapi.yaml
Vernholio: Fixed.

Vernholio:
Waiting on approval from DBaaS team for the release of this functionality. (Parity concerns: UI issues are preventing release for the API.)

PR #939, opened 2024-05-07
Updates for DC Get Well Acct Availability

Vernholio:
### Changed

Updated account availability-related endpoints, to apply new functionality and bring them from beta to GA:

* Account Availability (GET /account/availability)
* Region Service Availability (GET /account/availability/{id})

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -548,16 +548,17 @@ paths:
     get:
       tags:
       - account
+      parameters:
+      - $ref: '#/components/parameters/pageOffset'
+      - $ref: '#/components/parameters/pageSize'
       summary: Account Availability
       description: |
-        Returns service unavailability for all available regions.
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -548,16 +548,17 @@ paths:
...
+      - $ref: '#/components/parameters/pageSize'
       summary: Account Availability
       description: |
-        Returns service unavailability for all available regions.
+        Returns a paginated list of the services available to you, for all Linode regions.
 
-        Only authorized Users can access this endpoint.
-
-        *This endpoint is currently in **Beta**, and should not be used for production workloads.*
+        **Note**: Only authorized Users can access this endpoint.
```
Vernholio, line 558: What constitutes "authorized" for a user? I'm guessing an unrestricted user has access, but is there a specific permission that can be set for a user to give them access to this endpoint?

PR #940, opened 2024-05-08
Various requested fixes

Vernholio:
### Fixed

* Service Transfer Create (POST /account/service-transfers). Included VPC in list of services that a Linode can't use.
* VPC-related fixes:
  * Corrections to `nat_1._1` objects for proper support. 
  * Proper IP address formatting for VPC IP addresses.
  * IP Addresses List (GET /linode/instances/\{linodeId\}/ips). Corrected response example to show VPC's displayed as objects in an array and added `ipv4.vpc` attribute to `x-linode-cli-subtables`.
* Domain Clone (POST /domains/\{domainId\}/clone). Updated regex examples to proper format.
* IP Address List (GET /linode/instances/\{linodeId\}/ips). Replaced `requestBody` with query string.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -892,7 +892,18 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ChildAccount'
+                type: object
+                properties:
+                  data:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/ChildAccount'
+                  page:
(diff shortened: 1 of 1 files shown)


alsaihn (approved):

zliang-akamai:
Since `IPAddressesVPCListResponse` no longer represents the actual list API response and is now being a VPC IP object, can we consider rename it? How does `IPAddressVPC` sound?

Otherwise looks good to me. Thank you so much for fixing the issues in the doc!

PR #949, opened 2024-05-15
Linode disk encryption

wildmanonline:


wildmanonline:
@sandrajsalomone can you review / approve this PR if the changes are what you expect. They should be, since I used your same branch.

openapi.yaml
```diff
@@ -6311,6 +6311,10 @@ paths:
       summary: Image Create
       description: |
         Captures a private gold-master Image from a Linode Disk.
+
+        {{< note type="warning" title="Images Are Not Encrypted">}}
+        Images are not encrypted even when they are taken from an encrypted disk. When a compute instance is rebuilt from an image, and if the instance has disk encryption enabled, the disk is automatically encrypted.
+        {{< /note >}}
```
Vernholio, line 6317: For consistency with other "Note" entries in this file, can we change all of these note entries to:

\*\*Note\*\*: \<Note text\>.

I'm pretty sure this will be necessary for the migration... I doubt the Readme publishing tooling has a way to reference the short code layout.

sandrajsalomone (approved):

openapi.yaml
sandrajsalomone: I had a jira and feedback to use a warning, but if readme api ref doesn't support we can change these warnings.

PR #953, opened 2024-05-28
WIP: API docs for LKE NodePool labels and taints

mrogers950:
This PR adds docs for the upcoming LKE API feature of adding Kubernetes labels and taints to node pools.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -10978,11 +10978,13 @@ paths:
       - Linode Kubernetes Engine (LKE)
       summary: Node Pool Update
       description: |
-        Updates a Node Pool's count and autoscaler configuration.
+        Updates a Node Pool's count, labels and taints, and autoscaler configuration.
 
         Linodes will be created or deleted to match changes to the Node Pool's count.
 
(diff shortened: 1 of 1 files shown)


mrogers950:
Just tagging as WIP for now to ensure we don't merge before the feature

openapi.yaml
```diff
@@ -25694,6 +25707,99 @@ components:
...
+
+            * Specifying an empty dictionary value will remove all previously set labels.
+          additionalProperties:
+            type: object
+            properties:
+              key:
+                type: string
+                minLength: 1
+                maxLength: 63
+                pattern: '^([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]${1,63}'
```
mrogers950, line 25723: note for reviewers: I'm not sure about including the pattern as it's specified because it doesn't take into account the third bullet point in the description below, which changes the length and allowed pattern on the prefix.

wildmanonline:
Closing this in favor of #954

openapi.yaml
sandrajsalomone: The syntax {1,63} placed after $, I'm not sure that's correct..

Can I remove `${1,63}` from the syntax?

PR #956, opened 2024-08-02
New openapi.yaml file for 08-02-2024

Vernholio:
New instance of the openapi.yaml file for integration with the CLI.

PR #283, opened 2020-07-09
Add `disks` field to LKENodePoolRequestBody Schema

leslitagordita:
* Add disks field to LKENodePoolRequestBody schema
* This update affects the following endpoints:
  * POST /lke/clusters
  * POST /lke/clusters/{clusterId}/pools
  * Updated Shell and CLI examples for each endpoint.

PR #961, opened 2024-09-09
Fix example disk type values for LKE (`ext-4` vs `ext4`)

mabojars:
The Linode API supports either `ext4` or `raw` for the disk `type` value, in LKE routes. The OpenAPI spec examples suggest `ext-4` (with the `-`), which is incorrect. Requests using the `ext-4` value yield a validation error response from the Linode API.

Example:

```bash
curl --request POST \
     --url https://api.linode.com/v4/lke/clusters \
     --header 'accept: application/json' \
     --header "authorization: Bearer YEEEEHAW!" \
     --header 'content-type: application/json' \
     --data '
{
  "k8s_version": "1.30",
  "label": "lkecluster12345",
  "node_pools": [
    {
      "autoscaler": {
        "enabled": true,
        "max": 12,
        "min": 3
      },
      "count": 6,
      "disks": [
        {
          "size": 1024,
          "type": "ext-4"
        }
      ],
      "labels": {
        "example.com/my-app": "teams"
      },
      "tags": [
        "example tag",
        "another example"
      ],
      "taints": [
        {
          "effect": "NoSchedule",
          "key": "example.com/my-app",
          "value": "teamA"
        },
        {
          "effect": "NoExecute",
          "key": "myapp.io/team",
          "value": "teamB"
        }
      ],
      "type": "g6-standard-4"
    }
  ],
  "region": "us-central",
  "tags": [
    "ecomm",
    "blogs"
  ]
}
'

{"errors": [{"reason": "Must be one of raw, ext4", "field": "node_pools[0].disks[0].type"}]}
```

It appears the `ext-4` vs `ext4` discrepancy was introduced in [linode-api-docs/pull/283 ](https://github.com/linode/linode-api-docs/pull/283)(1 match for `ext-4`), and then disseminated in [linode-api-docs/pull/956 ](https://github.com/linode/linode-api-docs/pull/956)(6 matches for `ext-4`).

This PR replaces all occurrences of `ext-4` with the valid `ext4`, for LKE routes.


mabojars:
@Vernholio will take over introducing these changes from here - thanks!

PR #971, opened 2024-10-22
api-v4.189.2

Vernholio:
List images operation: Fix for the `x-linode-cli-display` placement for the `region` object.


openapi.yaml
```diff
@@ -29228,6 +29228,7 @@ paths:
                                 description: The unique identifier for the core compute region where this image is stored.
                                 example: us-iad
                                 type: string
+                                x-linode-cli-display: 11
```
yec-akamai, line 29231: Can you remove all the `x-linode-cli-display: 11`? It caused the output only display `region` instead of both `region` and `status`.

yec-akamai (approved):

